PR #26, opened 2026-04-19
test: reproduce streaming PdfiumStripSource rotation panic (libviprs#41)

spdrman:
## Summary

Adds `tests/streaming_pdfium_repro.rs` reproducing the slice OOB panic from [libviprs#41](https://github.com/libviprs/libviprs/issues/41) on the existing `blueprint.pdf` fixture.

## Tests

- `pdfium_streaming_blueprint_1mb_budget_replicates_pr96` — `#[ignore]` (intentionally panics)
- `pdfium_streaming_blueprint_2mb_budget_replicates_pr96_followup` — `#[ignore]` (intentionally panics; same panic, different memory budget)
- `pdfium_streaming_blueprint_actual_dims_should_pass` — control test, passes when actual rendered dims are passed instead of `pdf_info`-predicted
- `diagnostic_predicted_vs_actual_dims` — prints predicted vs actual dimensions

Run the panicking cases with:

```
cargo test --test streaming_pdfium_repro -- --ignored
```

Once libviprs#41 is fixed, drop the `#[ignore]` attributes so they become regression guards.

## Why two commits

The first commit (`fix: update tests for libviprs API drift`) updates `tests/streaming_engine.rs` and `tests/blank_tile_strategy.rs` for `pub`-export reorganisation in libviprs (`compute_strip_height` / `estimate_streaming_memory` now live in `libviprs::streaming`; `BLANK_TILE_MARKER` / `is_blank_tile` moved into `libviprs::sink` / `libviprs::engine`). Without this, `cargo clippy --all-targets` (and the pre-commit hook) fails on plain `main`.

The second commit adds the reproduction test itself.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -16,7 +16,7 @@ checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
 dependencies = [
  "cfg-if",
  "cipher",
- "cpufeatures",
+ "cpufeatures 0.2.17",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


spdrman:
Superseded by the matrix-path test gate in libviprs-tests#27, which is now the live regression guard for the rotation panic fixed in libviprs#43. The repro tests in this PR were marked #[ignore] and would not act as a guard once landed; the matrix-path tests in #27 do.

Issue #44, opened 2026-04-26
Migrate invariant tests off synthetic raster/PDF generators onto canonical fixtures


spdrman:
## Context

#43 lands the canonical input fixture set (`canonical.pdf`, `canonical_solid_white.pdf`, `canonical_rotated_{90,180}.pdf`, `canonical_input.png`) plus the offline generator (`tests/gen_canonicals.rs`). It does **not** convert any existing tests — the canonicals are unconsumed today.

This issue tracks the follow-up work to migrate ~30 integration tests off their per-file synthetic generators (`gradient_raster`, `solid_raster`, etc.) onto the canonical inputs, eliminating ~38 distinct ad-hoc helpers and one runtime PDF mutation path.

The work is split into bucketed PRs so each one is reviewable; bucket order is by risk/coverage, not by test count.

### Bucket A — pixel-fidelity tests (HIGHER risk, do these first)

Tests that compare libviprs output bytes to vips fixtures or to a known-good raster. Drift in their input matters: a moved pixel changes the assertion target.

- [ ] `tests/pyramid_determinism.rs`
- [ ] `tests/streaming_engine.rs`
- [ ] `tests/pdfium_pyramid_engine_parity.rs`

For each: replace the synthesised input with the appropriate canonical fixture, regenerate the corresponding `tests/fixtures/blueprint_*_expected/` tile set if the input changes (or, if a canonical isn't a drop-in, leave the test alone and document why in the PR).

### Bucket B — invariant tests (LOWER risk, ~25 files)

Tests that assert engine determinism, builder commutativity, resume / retry / sink contracts. They need *some* multi-channel non-uniform input but don't compare against a checked-in expected output. Swap `gradient_raster(w, h)` calls for a load of `canonical_input.png` (with a thin `tests/common/fixtures.rs` helper, if it makes the diff smaller).

- [ ] `tests/builder_*.rs` (composition, commutativity, etc.)
- [ ] `tests/phase3_*.rs` (sink, archive, dedupe paths)
- [ ] Any other test file under `tests/` that constructs its own raster via a `gradient_*` / `solid_*` helper and doesn't compare to a fixture.

### Bucket C — runtime PDF mutation (eliminate)

[...]

PR #46, opened 2026-04-26
test(fixtures): add CMYK canonicals + document stress.rs exception

spdrman:
Stacked on #45. Closes Buckets C and D from #44.

## Bucket C — eliminate runtime CMYK PDF construction

`tests/pdf_cmyk.rs` previously hand-built CMYK PDFs at test time (raw byte layout + xref tables, ~150 LOC of inline PDF assembly). This PR extends the offline canonical generator to produce three checked-in CMYK fixtures and rewrites `pdf_cmyk.rs` to load them:

| File | Purpose |
|---|---|
| `canonical_cmyk.pdf` | 16×16 DeviceCMYK FlateDecode gradient. Drives the end-to-end `extract_page_image` test. |
| `canonical_cmyk_black_1x1.pdf` | 1×1 K=255 → asserts CMYK→RGB full-black conversion. |
| `canonical_cmyk_cyan_1x1.pdf` | 1×1 C=255 → asserts pure-cyan conversion. |

Total disk footprint: ~2 KB.

The generator pre-compresses the CMYK bytes with flate2 and explicitly sets `/Filter /FlateDecode` on the image stream dict rather than relying on `Document::compress()`. That whole-doc pass skips small streams (the 1×1 fixtures end up uncompressed otherwise) and libviprs's CMYK extraction path keys on FlateDecode being present. The generator also reload-checks the round-tripped image XObject dimensions, mirroring the `/Rotate` self-check on the existing canonicals.

`pdf_cmyk.rs` shrinks from 200 → 67 LOC and no longer constructs PDFs at test time.

## Bucket D — document the stress.rs synthetic-retention rationale

`tests/stress.rs` retains its `synthetic_raster` helper. Module doc comment now spells out why:

1. The stress paths run at 10K × 10K and 4K × 4K, well above the 256 × 256 canonical input. Aggressive nearest-neighbour upscaling produces blocky regions that don't exercise downscale-averaging the way a unique-per-pixel pattern does.
2. Checking in a 10K × 10K RGB8 PNG (~300 MB) just to feed these tests would dominate the repo's disk footprint.
3. Every test in the file is `#[ignore]`d by default, so the synthetic input never feeds an automated comparison.

## Net effect across the migration (PRs #43 + #45 + this one)

After this PR merges, **every regular test in the suite loads its inputs from `tests/fixtures/`**. The only synthetic raster construction left is in:

- `tests/gen_canonicals.rs` and `tests/gen_blank_tile_fixtures.rs` — the fixture *generators* themselves, by design.
- `tests/stress.rs` — the documented Bucket D exception (all `#[ignore]`d).

## Test plan

- [x] `cargo test --test gen_canonicals -- --ignored` produces deterministic byte-stable CMYK fixtures.
- [x] `cargo test --test pdf_cmyk` — 3/3 passing against the new fixtures.
- [x] `cargo fmt --check` + `cargo clippy --all-targets -- -D warnings` clean.
- [x] Full Docker test suite via the pre-push hook: clean.


spdrman:
Superseded by a fresh PR cherry-picked onto current main. Branch protection blocks pushing the merge-resolved state to this branch directly, so a clean replacement is the path forward.

PR #47, opened 2026-04-26
tools(install-hooks): mirror each repo's CI clippy matrix exactly

spdrman:
## Summary

The pre-commit hook used to run a single `cargo clippy --all-targets` across every repo, which left CI-only feature matrices uncovered. Specifically, libviprs's CI runs **both** default-features and `--features pdfium` clippy passes; the local hook only ran the default-features pass. Lints that fired only with `--features pdfium` were invisible at commit time and broke CI on push, forcing a correction loop (most recently: `clippy::doc_lazy_continuation`, `clippy::large_enum_variant`, and `clippy::derivable_impls` on libviprs#71).

This PR refactors `install-hooks.sh` so each repo gets a hook that mirrors its own `.github/workflows/ci.yml` Check & Lint cargo invocations verbatim. Per-repo command lists are declared at the top of the script as named arrays — keep those in lockstep with the YAML and re-run the script when CI changes.

## What's in the new hooks

| Repo | Cargo lint commands the hook runs |
|---|---|
| `libviprs` | fmt + clippy default + clippy `--features pdfium` |
| `libviprs-cli` | fmt + clippy default |
| `libviprs-tests` | fmt + clippy default |

(Each repo's commands are exactly the cargo lines under that repo's CI Check & Lint job.)

## Other cleanups in the same pass

- Drop `2>/dev/null` on cargo invocations: when a lint fires we *want* to see the error wall immediately, not silence it and force a manual rerun.
- Factor `write_pre_commit` / `write_pre_push` into helpers so the repo loop reads as orchestration rather than nested heredoc.
- Per-step echo + per-step failure message: when a check fails, the hook prints the exact CI line the commit would have flunked.
- Pre-push hook (Docker test suite) is unchanged — its CI parity was already correct.

## Test plan

- [x] Re-run `./tools/install-hooks.sh` from workspace root: emits hooks for all three repos with their respective command lists.
- [x] Confirmed via `grep` that each generated `pre-commit` hook contains the right cargo invocations for its repo (libviprs has both default + pdfium; the others have just default).
- [x] On libviprs, the new hook caught the two `--features pdfium`-only lints (`doc_lazy_continuation` + `large_enum_variant`) at commit time before the matching CI run would have failed.
- [x] Full Docker test suite passes via the (unchanged) pre-push hook.


spdrman:
Superseded by a new PR that bundles this work with the pdfium-render fork-scoped pre-commit. Branch protection on libviprs-tests blocks direct pushes to feature branches with open PRs, so combining the two install-hooks changes required a fresh branch.

PR #48, opened 2026-04-26
tools(install-hooks): mirror CI per-repo + add fork-scoped pdfium-render hook

spdrman:
Supersedes #47 (closed). Branch protection blocks direct pushes to existing feature branches even when their PR is open, so combining the original CI-mirror change with the pdfium-render fork-scoped hook required a fresh branch.

Stacked on #46. Two related commits:

## 1. `d8e28fb` — Mirror each repo's CI Check & Lint job in pre-commit

Originally PR #47. The pre-commit hook used to run a single `cargo clippy --all-targets` across every repo, leaving CI-only feature matrices uncovered. libviprs's CI runs **both** default-features and `--features pdfium` clippy passes; the local hook only ran the default-features pass. Lints that fired only with `--features pdfium` were invisible at commit time and broke CI on push (`clippy::doc_lazy_continuation`, `clippy::large_enum_variant`, `clippy::derivable_impls` on libviprs#71).

`install-hooks.sh` now emits a per-repo hook whose cargo command list mirrors that repo's `.github/workflows/ci.yml` Check & Lint job verbatim. Per-repo command lists are declared at the top of the script as named arrays — keep those in lockstep with the YAML and re-run the script when CI changes.

| Repo | Cargo lint commands the hook runs |
|---|---|
| `libviprs` | fmt + clippy default + clippy `--features pdfium` |
| `libviprs-cli` | fmt + clippy default |
| `libviprs-tests` | fmt + clippy default |

Other cleanups in this commit: per-step echo + per-step failure messages so a failing hook prints the exact CI line it would have flunked; drop `2>/dev/null` so cargo errors are visible; factor the heredoc generation into helpers.

## 2. `38c4865` — Fork-scoped pre-commit for pdfium-render

Upstream pdfium-render carries hundreds of pre-existing clippy lints (~470 missing-safety-doc, ~22 doc-nested-refdefs, useless-vec, etc.), so a blanket `cargo clippy -- -D warnings` would block every commit including merges from upstream. The pdfium-render hook this script now installs runs clippy whole-tree but filters its JSON output to only those lints whose primary span sits on a line *this fork has actually changed* relative to its base on `upstream/master` (per `git diff` hunks).

Net: strict on lints we introduce, silent on the ambient debt we don't own. False-positive escape hatch documented inline — if a lint sits on a line we changed but the underlying issue is upstream's, revert our edit on that line or `git commit --no-verify`.

Implementation: hook body uses python3 to parse `git diff --unified=0` hunk headers and `cargo clippy --message-format=json` together. The repo loop case-switches: pdfium-render gets the scoped pre-commit only (no Docker pre-push — that repo has no `run-tests.sh` integration); everything else keeps the existing CI-mirror pre-commit + Docker pre-push.

Verified by running the installed hook on the two open libviprs PR branches against ajrcarey/pdfium-render (#249 start-xy-passthrough, #252 auto-apply-rotation): both pass clean. The only fork-introduced lint that the original whole-tree-no-scope check found was the `unnecessary_cast` in #249's strip-stitching test, which had already been fixed before this hook landed.

## Test plan

- [x] `./tools/install-hooks.sh` from workspace root: emits hooks for all four repos with their respective configurations.
- [x] Confirmed via `grep` that each generated `pre-commit` hook contains the right cargo invocations for its repo.
- [x] On libviprs, the new hook caught the two `--features pdfium`-only lints (`doc_lazy_continuation` + `large_enum_variant`) at commit time before the matching CI run would have failed.
- [x] On pdfium-render, the scoped hook reports clean for both libviprs/start-xy-passthrough and libviprs/auto-apply-rotation.
- [x] Full Docker test suite passes via the (unchanged) pre-push hook on the libviprs/libviprs-cli/libviprs-tests trio.


spdrman:
Superseded by a fresh PR cherry-picked onto current main (CMYK content already landed via #49, so the install-hooks commits stand on their own now).

PR #52, opened 2026-04-27
test(pdfium): TDD reds for perf/concurrency/regression + DRY helpers

spdrman:
Replaces #42 (closed). Same content as #42's `036af3f` cherry-picked onto current `main`, with conflict resolutions where the now-landed fixture stack (#43, #45, #49, #50) and streaming-mode tests (#51) had introduced overlapping shared infrastructure.

## What this adds

- **`tests/common/mod.rs`** — pdfium-specific shared helpers (fixture path constants, region-mean tolerances, channel-mean assertion helpers, `crop_top_left`). The previously-existing `tests/common/mod.rs` from #45's canonicalisation work re-exports `pub mod fixtures;`; the pdfium helpers now sit in a `mod pdfium_helpers` block gated by `#[cfg(feature = "pdfium")]` so the file compiles cleanly under both feature configurations.
- **TDD-red tests** for libviprs streaming-mode scenarios:
  - `pdfium_streaming_concurrent.rs` — concurrent `render_strip` calls from multiple threads.
  - `pdfium_streaming_determinism.rs` — same input → identical output across runs.
  - `pdfium_streaming_perf_smoke.rs` — wall-clock floor so streaming-mode regressions get caught.
  - `pdfium_streaming_property.rs` — proptest-style invariants over strip y_offset / height.
  - `pdfium_streaming_regression_rotate.rs` — pin the prior 180°-off `/Rotate 90/270` bug.
- **DRY** — `pdfium_streaming_render.rs` / `_rotation_matrix.rs` / `_memory.rs` / `_synthetic_rotations.rs` deduped to consume `common::*` instead of carrying their own fixture constants and assertion helpers.

## Conflict resolutions during cherry-pick

- **`tests/common/mod.rs`**: combined the canonical-fixtures stub from #45 with the pdfium helpers from this PR. Single file, two responsibilities, gated cleanly.
- **`tests/pdfium_streaming_synthetic_rotations.rs`**: kept this PR's `PageRotation::try_from_degrees` round-trip assertion (the more type-safe form) over the interim `as_degrees()` patch I'd applied to #51's branch.
- **`tests/pdfium_streaming_rotation_matrix.rs`**: import block merged so the `mod common;` line and the `PageRotation` import both make it through.

## Test plan

- [x] `cargo build --tests --features pdfium` clean against current libviprs/main.
- [x] Docker pre-push test suite passes.
- [x] All five new TDD-red test files compile against the post-libviprs#71 streaming API.

PR #53, opened 2026-04-27
test(pdfium): TDD reds for perf/concurrency/regression + DRY helpers

spdrman:
Replaces #52 (closed). Single squashed commit on top of current main.

Same content as #52 plus the `#[allow(unused_imports)]` on the `pub use pdfium_helpers::*;` re-export — that line trips CI's `-D warnings` on tests that consume `mod common;` for the canonical-fixture loader without touching any pdfium-specific helper. The fix was committed on top of the previous branch but branch protection rejected the push, so this PR squashes the original work + the fix into one clean commit.

## What this adds

- **`tests/common/mod.rs`** — pdfium-specific shared helpers (fixture path constants, region-mean tolerances, channel-mean assertion helpers, `crop_top_left`). Coexists with the canonical-fixture stub from #45 — pdfium helpers are gated behind `#[cfg(feature = "pdfium")]` and the re-export at the root carries `#[allow(unused_imports)]` so non-pdfium consumers of `common::fixtures::*` compile cleanly.
- **TDD-red tests** for libviprs streaming-mode scenarios:
  - `pdfium_streaming_concurrent.rs` — concurrent `render_strip` calls from multiple threads.
  - `pdfium_streaming_determinism.rs` — same input → identical output across runs.
  - `pdfium_streaming_perf_smoke.rs` — wall-clock floor so streaming-mode regressions get caught.
  - `pdfium_streaming_property.rs` — proptest-style invariants over strip y_offset / height.
  - `pdfium_streaming_regression_rotate.rs` — pin the prior 180°-off `/Rotate 90/270` bug.
- **DRY** — `pdfium_streaming_render.rs` / `_rotation_matrix.rs` / `_memory.rs` / `_synthetic_rotations.rs` deduped to consume `common::*` instead of carrying their own fixture constants and assertion helpers.

## Test plan

- [x] `cargo build --tests --features pdfium` clean.
- [x] `cargo build --tests` (default features) clean — the `unused_imports` lint doesn't fire on consumers that don't use pdfium helpers.
- [x] Docker pre-push test suite passes.

### DIFF
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -2,11 +2,177 @@
 //!
 //! Each `tests/*.rs` file is compiled as a standalone integration-test
 //! binary, so anything they share has to live here under `tests/common/`
-//! and be pulled in with `mod common;` at the top of the test file. The
-//! `#[allow(dead_code)]` on the helpers below is intentional — not every
+//! and be pulled in with `mod common;` at the top of the test file.
+//! Cargo treats `tests/common/mod.rs` (directory module) specially — it
+//! isn't compiled as its own integration target.
+//!
+//! The canonical-fixture loader sits in [`fixtures`]. Below it: PDF /
+//! pdfium-specific shared helpers for the streaming-mode integration
+//! tests. The pdfium helpers are gated behind the `pdfium` feature so
+//! the module compiles cleanly with the default feature set.
+//!
+//! `#[allow(dead_code)]` on the helpers is intentional — not every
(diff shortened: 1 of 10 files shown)


spdrman:
Superseded — also had a per-file unused-import (FIXTURE_MIX in pdfium_streaming_render.rs) caught by CI's RUSTFLAGS=-Dwarnings that I missed locally. Replacement PR amends the fix in.

Issue #87, opened 2026-07-13
Enable 6 ported tests gated on non-codec core features (DeepZoom layouts, ZipSink, blank-tile skip, region, PDF background)


spdrman:
During the #77 codec-cell enablement (PR #85), 6 ported tests turned out to reference core symbols OUTSIDE the codec surface, so they were left as compile-safe `#[ignore]` stubs (their spec is preserved in the doc comments) rather than blocking the 95 real codec tests. Each wants a core follow-up in libviprs/libviprs before it can run:

- `Layout::Zoomify` and `Layout::Iiif` (DeepZoom layout variants on the pyramid sink)
- `ZipSink` (pyramid output to a zip container)
- `EngineConfig::skip_blanks` (blank-tile skipping toggle)
- `generate_pyramid_region` (render a sub-region of the pyramid)
- `extract_page_image_with_background` (PDF page render with a background fill)

These are pyramid-engine / PDF-render features, not part of the ported codec surface (#77). Done means: land the core symbol, then un-ignore the corresponding ported test (drop the stub) and bump COUNTERPART_REV.

Separately, ported_foreign still carries the permanently-deferred external-codec tests (HEIF/AVIF/JP2K/JXL/WebP-encode, magickload/save, animated GIF, OJPEG/OME-multipage TIFF, UHDR gainmap, JPEG EXIF/ICC chroma roundtrips), which assert the typed Unsupported contract and stay ignored by design (no pure-Rust path).

spdrman:
All six sub-issues landed and merged to main:

- #88 `Layout::Zoomify` → #94
- #89 `Layout::Iiif` → #95
- #90 `ZipSink` → #96
- #91 `EngineConfig::skip_blanks` → #98
- #92 `generate_pyramid_region` → #97
- #93 `extract_page_image_with_background` → #100 (superseded #99)

Each ported test was un-ignored and driven as a vips-differential golden test (libviprs output byte-matched against a checked-in `vips`-generated reference), `COUNTERPART_REV` bumped to `6f094b0e`, and verified green under the local pre-commit gate + the full `ported_foreign` cell (85 passed / 0 failed; pdf test green under pdfium). A 4-lens expert review (architecture / rust-systems / safety / API) is recorded on each PR; core-API follow-ups are being filed against libviprs/libviprs.

PR #95, opened 2026-07-18
test(#89): IIIF DeepZoom layout ported test (vips-differential)

### LINKED ISSUE #89: Un-ignore test_dz_layout_iiif (IIIF DeepZoom layout)
Part of #87.

## Feature
Enable the ported test `test_dz_layout_iiif` in `tests/ported_foreign.rs`, which is currently a compile-safe `#[ignore]` stub whose spec is preserved in its doc comment.

`Layout::Iiif` shipped in libviprs/libviprs core PR #318 (on core main at `6f094b0e7955d29a6fdbc2596cfdd41189a4766f`). IIIF writes an info.json sidecar (width/height) plus region directories, no .dzi.

## Done means
- Drop the deferred `#[ignore]` stub and restore a real test body exercising `Layout::Iiif`, matching the doc-comment spec and the surrounding `test_dz_*`/`test_pdf_*` style.

spdrman:
Part of #87. Closes #89.

- Core symbol `Layout::Iiif` landed in libviprs core PR #318 (main @ `6f094b0e`); this PR bumps `COUNTERPART_REV` to that revision.
- Compares every {region}/{size},/0/default.png tile + info.json against a vips `--layout iiif` reference.


spdrman:
## Consolidated expert review

### PR-scope
- No material issues. The `info.json` width/height assertion plus the region-directory / no-`.dzi` structural checks correctly pin the IIIF contract; the tile comparison covers the pixels.

- **Low (API):** the emitted `info.json` hardcodes `"@id": "https://example.com/iiif"`. IIIF ties `@id` to the serving location, so every caller must string-rewrite it; there is no typed setter. Consider a `base_url` parameter or a documented post-process hook so the placeholder can't ship by accident.

Approve.

PR #96, opened 2026-07-18
test(#90): ZipSink archive tile sink ported test (vips-differential)

### LINKED ISSUE #90: Un-ignore test_dz_zip (ZipSink archive tile sink)
Part of #87.

## Feature
Enable the ported test `test_dz_zip` in `tests/ported_foreign.rs`, which is currently a compile-safe `#[ignore]` stub whose spec is preserved in its doc comment.

`ZipSink` shipped in libviprs/libviprs core PR #316 (on core main at `6f094b0e7955d29a6fdbc2596cfdd41189a4766f`). ZipSink::new/try_new over PackfileSink(Zip). Needs the packfile feature; PR also wires a ported_tests+packfile run into tools/run_ported_cells.sh so CI exercises it.

## Done means
- Drop the deferred `#[ignore]` stub and restore a real test body exercising `ZipSink`, matching the doc-comment spec and the surrounding `test_dz_*`/`test_pdf_*` style.
- Gating: `feature = "ported_tests" + feature = "packfile"`.

spdrman:
Part of #87. Closes #90.

- Core symbol `ZipSink` landed in libviprs core PR #316 (main @ `6f094b0e`); this PR bumps `COUNTERPART_REV` to that revision.
- Gated on `ported_tests`+`packfile`; extracts the archive and compares its DeepZoom `_files` subtree to a vips dz reference. Also wires a `ported_tests packfile` clippy+test pass into `tools/run_ported_cells.sh`.


spdrman:
## Consolidated expert review

### PR-scope
- **Suggestion:** comparing the archive's DeepZoom `tiles_files/` subtree against the FS reference is the right call, but it couples the test to the archive stem (`tiles` from `tiles.zip`). A short comment stating that dependency (already partly present) is enough; no change required.
- **Mandatory (Architecture):** `zip_expected.dzi` is committed as a fixture but never read — the comparison is tile-only. Remove the unused `.dzi` (its vips multi-line format wouldn't match core's single-line manifest anyway), or add a semantic `.dzi` assertion. **This PR removes it.**
- The `tools/run_ported_cells.sh` wiring (a `ported_tests packfile` clippy + test pass) is correct and necessary — no other CI job exercises that feature combination.

- **Medium (Systems/API):** `PackfileSink::write_tile` writes **every tile twice** into the archive — once under `<stem>_files/<level>/<x>_<y>.<ext>` and again under `<stem>/…`. That roughly doubles archive size and entry count for a compatibility convenience; worth gating behind an option or dropping one layout.
- **Low (API):** three divergent `new` contracts across the sink family — `ZipSink::new` panics on I/O, `ZipSink::try_new` and `PackfileSink::new` return `Result`. Consider making the fallible form the canonical `new`.

Approve after the fixture removal; core items are follow-ups.

PR #97, opened 2026-07-18
test(#92): generate_pyramid_region ported test (vips-differential)

### LINKED ISSUE #92: Un-ignore test_dz_region (generate_pyramid_region)
Part of #87.

## Feature
Enable the ported test `test_dz_region` in `tests/ported_foreign.rs`, which is currently a compile-safe `#[ignore]` stub whose spec is preserved in its doc comment.

`generate_pyramid_region` shipped in libviprs/libviprs core PR #317 (on core main at `6f094b0e7955d29a6fdbc2596cfdd41189a4766f`). Crop-then-pyramid: plan is sized to the region (not the source); a dimension mismatch is a typed error.

## Done means
- Drop the deferred `#[ignore]` stub and restore a real test body exercising `generate_pyramid_region`, matching the doc-comment spec and the surrounding `test_dz_*`/`test_pdf_*` style.

spdrman:
Part of #87. Closes #92.

- Core symbol `generate_pyramid_region` landed in libviprs core PR #317 (main @ `6f094b0e`); this PR bumps `COUNTERPART_REV` to that revision.
- Crop-then-pyramid matches vips `crop`+`dzsave` of the same sub-region.


spdrman:
## Consolidated expert review

### PR-scope
- **Mandatory (Architecture):** `region_expected.dzi` is committed but never read (tile-only comparison). Remove it. **This PR removes it.**
- Good: the test pins the crop-then-pyramid contract by sizing the plan to the region (not the source) and asserting `tiles_produced == plan.total_tile_count()` plus the top-level tile bound — this is exactly the invariant that would break if the crop were mis-applied.

- **Low (API):** `generate_pyramid_region` is an eight-argument free function with no `EngineBuilder` equivalent, so region generation can't compose with the builder surface (resume, concurrency, observers). A `Region` newtype or a builder method would cut the arg count and unify the paths.
- **Low (API/Safety):** a plan-vs-region dimension mismatch is reported as `EngineError::PlanSourceMismatch` whose fields are named `source_*` — misleading when the mismatch is against the *region*, not the source. A dedicated variant (or renamed fields) would read truthfully.

Approve after the fixture removal.

PR #98, opened 2026-07-18
test(#91): EngineConfig::skip_blanks ported test (vips-differential)

### LINKED ISSUE #91: Un-ignore test_dz_skip_blanks (EngineConfig::skip_blanks)
Part of #87.

## Feature
Enable the ported test `test_dz_skip_blanks` in `tests/ported_foreign.rs`, which is currently a compile-safe `#[ignore]` stub whose spec is preserved in its doc comment.

`EngineConfig::skip_blanks` shipped in libviprs/libviprs core PR #317 (on core main at `6f094b0e7955d29a6fdbc2596cfdd41189a4766f`). skip_blanks(true) drops fully-uniform tiles from the output, so a mostly-blank source yields strictly fewer tiles.

## Done means
- Drop the deferred `#[ignore]` stub and restore a real test body exercising `EngineConfig::skip_blanks`, matching the doc-comment spec and the surrounding `test_dz_*`/`test_pdf_*` style.

spdrman:
Part of #87. Closes #91.

- Core symbol `EngineConfig::skip_blanks` landed in libviprs core PR #317 (main @ `6f094b0e`); this PR bumps `COUNTERPART_REV` to that revision.
- Mostly-white source with three pure-white quadrants; libviprs `skip_blanks(true)` drops exactly the same tiles as vips `--skip-blanks 0`.


spdrman:
## Consolidated expert review

### PR-scope
- **Mandatory (Architecture):** `skip_expected.dzi` is committed but never read. Remove it. **This PR removes it.**
- Strong test design: the source is constructed so the blank tiles are *exactly* pure white, which makes libviprs' exact-uniformity skip and vips' `--skip-blanks 0` agree deterministically (they kept the same 10 of 13 tiles). The `result.tiles_skipped > 0` guard prevents a vacuous pass.

- **Low (API):** `EngineConfig::skip_blanks(bool)` is the only non-`with_`-prefixed setter on the config, it shares a name with the public `skip_blanks` field, and there is no `EngineBuilder::with_skip_blanks` mirror — inconsistent with the rest of the builder surface. It also overlaps `BlankTileStrategy` and is documented as monolithic-engine-only; the relationship should be spelled out or unified.

Approve after the fixture removal.

PR #99, opened 2026-07-18
test(#93): PDF background render ported test (vips-differential)

### LINKED ISSUE #93: Un-ignore test_pdf_background (extract_page_image_with_background)
Part of #87.

## Feature
Enable the ported test `test_pdf_background` in `tests/ported_foreign.rs`, which is currently a compile-safe `#[ignore]` stub whose spec is preserved in its doc comment.

`extract_page_image_with_background` shipped in libviprs/libviprs core PR #319 (on core main at `6f094b0e7955d29a6fdbc2596cfdd41189a4766f`). Requires the pdfium feature (returns typed Unsupported without it). Test mirrors the existing test_pdf_password pattern: #[cfg_attr(not(feature = "pdfium"), ignore)].

## Done means
- Drop the deferred `#[ignore]` stub and restore a real test body exercising `extract_page_image_with_background`, matching the doc-comment spec and the surrounding `test_dz_*`/`test_pdf_*` style.
- Gating: `feature = "ported_tests" + feature = "pdfium"`.

spdrman:
Part of #87. Closes #93.

- Core symbol `extract_page_image_with_background` landed in libviprs core PR #319 (main @ `6f094b0e`); this PR bumps `COUNTERPART_REV` to that revision.
- Pixel-match tolerance: **1.0 (observed 0)**.
- pdfium-gated (mirrors `test_pdf_password`); an empty-content PDF makes the whole render the background colour, matched against a vips `pdfload --background` reference. Also asserts typed errors for bad channel count / page 0.


spdrman:
## Consolidated expert review

**Verdict:** Sound, well-scoped test enablement. The vips-differential approach — run libviprs, byte-match the output against a checked-in `vips`-generated reference, no libvips at runtime — is the right pattern and mirrors the existing `blueprint_*_pyramid` tests. The shared `tests/common/dzsave_expected.rs` helper is clean, and pinning it byte-identical across the sibling PRs so `COUNTERPART_REV` and the helper auto-resolve on merge is a reasonable call. Tolerance `1.0` (observed diff `0`) is safe here: both sides are frozen bytes and libviprs tiling is deterministic, so it cannot flake on platform float drift the way a live re-render would.

### PR-scope
- Good: testing the background feature on an empty-content PDF (`canonical_solid_white.pdf`) makes the whole render the background colour, which isolates the feature from pdfium-vs-vips rasteriser divergence — that's why it lands an exact match instead of needing a wide tolerance. The typed-error assertions (wrong channel count, page 0) pin the validation contract, and the pdfium gate mirrors the established `test_pdf_password` pattern so it stays green in the libvips-free CI.
- **Suggestion (Safety):** the white-vs-red `max_diff > 0` check assumes the fixture is transparent; a one-line comment tying it to the empty-content-stream property would harden it against a future fixture swap.

- **Low (API):** `extract_page_image_with_background(path, page, &[f64])` takes a loosely-typed colour slice validated only at runtime (3 or 4 channels); a `Rgba`/`Color` type would move that to compile time. The free-function family also doesn't compose — there's no single entry point combining background + DPI + password.

Approve; core items are follow-ups.

spdrman:
Superseded by #100 — recreated off current main to resolve the shared-`use`-line conflict with #97 (the pull_request ruleset blocked rebasing this branch in place). Same test, same vips reference.

PR #100, opened 2026-07-18
test(#93): PDF background render ported test (vips-differential, supersedes #99)

### LINKED ISSUE #93: Un-ignore test_pdf_background (extract_page_image_with_background)

spdrman:
Part of #87. Closes #93. Supersedes #99 (which conflicted with #97 on a shared `use` line; the ruleset blocked rebasing it in place, so this is a clean recreate off current main).

Enables `test_pdf_background` as a **vips-differential golden test**: renders `canonical_solid_white.pdf` (empty content → whole render is the background colour, isolating the feature from pdfium-vs-vips rasteriser divergence) with libviprs and byte-matches it against a checked-in `vips pdfload --background` reference. Also asserts the typed-error contract (bad channel count, page 0). pdfium-gated, mirroring `test_pdf_password`.

Verified locally: pre-commit hooks (fmt + clippy across default/s3/packfile/tracing cells) + the pdfium pdf test green. `COUNTERPART_REV` already at `6f094b0e` on main.


spdrman:
## Consolidated expert review

**Verdict:** Sound, well-scoped test enablement. The vips-differential approach — run libviprs, byte-match the output against a checked-in `vips`-generated reference, no libvips at runtime — is the right pattern and mirrors the existing `blueprint_*_pyramid` tests. The shared `tests/common/dzsave_expected.rs` helper is clean, and pinning it byte-identical across the sibling PRs so `COUNTERPART_REV` and the helper auto-resolve on merge is a reasonable call. Tolerance `1.0` (observed diff `0`) is safe here: both sides are frozen bytes and libviprs tiling is deterministic, so it cannot flake on platform float drift the way a live re-render would.

### PR-scope
- Good: testing the background feature on an empty-content PDF (`canonical_solid_white.pdf`) makes the whole render the background colour, which isolates the feature from pdfium-vs-vips rasteriser divergence — that's why it lands an exact match instead of needing a wide tolerance. The typed-error assertions (wrong channel count, page 0) pin the validation contract, and the pdfium gate mirrors the established `test_pdf_password` pattern so it stays green in the libvips-free CI.
- **Suggestion (Safety):** the white-vs-red `max_diff > 0` check assumes the fixture is transparent; a one-line comment tying it to the empty-content-stream property would harden it against a future fixture swap.

- **Low (API):** `extract_page_image_with_background(path, page, &[f64])` takes a loosely-typed colour slice validated only at runtime (3 or 4 channels); a `Rgba`/`Color` type would move that to compile time. The free-function family also doesn't compose — there's no single entry point combining background + DPI + password.

Approve; core items are follow-ups.

PR #105, opened 2026-07-18
test(resample): Nearest-kernel RGBA reduce must return exact source pixels (repro #287/#288)

spdrman:
## RED repro: Nearest single-tap kernel corrupts semi-transparent RGB

Follow-up to the merged premultiply fix (libviprs/libviprs PR #337, issues #287/#288).

### The defect this pins
`reduce_axis` now premultiplies colour by alpha for **every** alpha format, unconditionally. That is correct for the averaging kernels, but the **Nearest** kernel is a single-tap *pick* (no averaging). Wrapping a pick in a premultiply -> un-premultiply round-trip through a **same-bit-depth integer** raster requantises the straight-alpha RGB of semi-transparent pixels:

```
premultiply:    c' = round(c * a / max)
un-premultiply: c'' = round(c' * max / a)   # lossy for low alpha
```

So the selected source pixel does **not** survive byte-exact.

### The invariant
A single-tap nearest reduce/resize of a semi-transparent RGBA image must return each selected source pixel **byte-identically** — exactly what `vips resize <rgba> --kernel nearest` guarantees. Stated directly as a pure invariant (no libvips at runtime, no fixture).

### Status: RED against current core main
All 5 cases fail; each output pixel equals the requantised round-trip of its source, e.g. `(30,220,140,3)` -> `(0,255,170,3)`, `(90,200,60,5)` -> `(102,204,51,5)`.

Covers `reduceh`, `reducev`, `reduce` (integer + fractional), and `resize` (nearest). Goes GREEN once the core skips premultiply for the Nearest single-tap kernel (core fix PR).

New file only: `tests/resample_nearest_alpha.rs`. No shared files or COUNTERPART_REV touched. [skip ci]

### DIFF
--- a/tests/resample_nearest_alpha.rs
+++ b/tests/resample_nearest_alpha.rs
@@ -1,0 +1,121 @@
+//! Regression repro for libviprs/libviprs follow-up to #287/#288 (PR #337).
+//!
+//! PR #337 made `reduce_axis` premultiply colour by alpha for *every* alpha
+//! format, unconditionally. That is correct for the averaging kernels (it stops
+//! transparent RGB bleeding into opaque neighbours), but the **Nearest** kernel
+//! is a single-tap pick: it does no averaging, so wrapping it in a
+//! premultiply -> un-premultiply round-trip can only corrupt the straight-alpha
+//! RGB of semi-transparent pixels. Because `premultiply_raster` premultiplies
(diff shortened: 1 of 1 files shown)


spdrman:
Superseded by #107, which lands this repro test (GREEN against the fixed core) together with the COUNTERPART_REV bump. RED-first provenance is preserved by this PR's history.

PR #106, opened 2026-07-18
test(composite): replicate #289 primary-path/output-tag gap (RED until core fix)

spdrman:
## What

Adds `tests/composite_289_primary_path.rs`, a pure public-API repro for the correctness gap an adversarial review found in the merged #289 fix (core PR #340).

## The gap (output-tag mismatch)

`composite2` writes a mixed-depth result at the genuine 16-bit scale whenever a genuine-16 (`Rgb16`/`Grey16`) input is present, so a half-opaque genuine-16 overlay's alpha correctly survives above the 255 ceiling on the **first** composite. But the result is tagged with the **base's** interpretation (`out.meta = self.meta`), not the genuine-16 compositing space it was actually written at.

Chaining that result back through `composite2` therefore re-reads it on the 255 scale and clamps the alpha back to the 255 ceiling — silently re-introducing #289 one hop downstream, and leaving saved metadata inconsistent with the samples.

## The test

`genuine_16bit_composite_result_survives_rechain`:
1. Composites a genuine-16 half-opaque overlay over an 8-bit base with `Source` → asserts the first-hop alpha lands near `0x8000` (already correct under #340).
2. Re-chains that result as the overlay of another `Source` composite → asserts the alpha still survives on the 16-bit scale.

Against the current core, step 2 observes `[10000, 20000, 30000, 255]` — the alpha is clamped to 255 (the re-introduced #289), so the test is **RED**.

## Status

RED until the core follow-up stamps the output interpretation to match its write-back scale (genuine-16 space when written at the 65535 scale). Draft; do not merge ahead of the core fix + counterpart bump.

Refs #289.

### DIFF
--- a/tests/composite_289_primary_path.rs
+++ b/tests/composite_289_primary_path.rs
@@ -1,0 +1,92 @@
+//! Follow-up repro for issue #289 (the mixed-depth compositing fix landed in
+//! libviprs/libviprs#340 / PR #340).
+//!
+//! `composite2` writes a mixed-depth result at the genuine 16-bit scale
+//! whenever a genuine-16 (`Rgb16`/`Grey16`) input is present — so a
+//! half-opaque genuine-16 overlay's alpha correctly survives above the 255
+//! ceiling on the FIRST composite. But the result is tagged with the *base's*
+//! interpretation (`out.meta = self.meta`), not the genuine-16 compositing
+//! space it was actually written at. Chaining that result back through
+//! `composite2` therefore re-reads it on the 255 scale and clamps the alpha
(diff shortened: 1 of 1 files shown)


spdrman:
Core fix: libviprs/libviprs#355 (tags the genuine-16 composite output to match its 65535 write-back scale). This repro turns GREEN once #355 merges and COUNTERPART_REV is bumped past it. The separable read-side genuine-vs-promoted ambiguity on the untagged primary construction path is tracked in libviprs/libviprs#356.

spdrman:
Superseded by #107, which lands this repro test (GREEN against the fixed core) together with the COUNTERPART_REV bump. RED-first provenance is preserved by this PR's history.

PR #108, opened 2026-07-18
test(#346): repro unguarded draw(&op) ink path + try_draw_mask float panic [skip ci]

spdrman:
## What

Red-first vips-differential repro for the PR #335 review follow-ups tracked by issue #346 (core `src/draw.rs`). Adds `tests/draw_op_path_validation.rs` with four cells over the public draw API (no libvips at runtime).

## Defects reproduced

1. **Unguarded `draw(&op)` / op-constructor ink path.** PR #335 added ink channel-count validation only to the `draw_*` / `try_draw_*` wrappers. The equally-public `Raster::draw(&op)` entry point plus the draw-op constructors (`Circle`, `Rectangle`, `Mask`, ...) stay an unguarded route to the exact #294 corruption: a wrong-width ink is cycled/truncated verbatim through `put_pixel`, silently painting channels from the wrong bytes (e.g. an `Rgba8` alpha band from the red byte). Covered across all three internal ink mechanisms: `put_pixel` (Rectangle, Circle) and `ink_pixel` (Mask).
2. **`try_draw_mask` panics on float rasters.** The *fallible* mask entry point still panics deep inside the unsigned-sample decode on an `RgbaF32` / `FloatF32` raster even with correct-width ink, instead of returning a typed `DrawError`.

## Status: RED (expected)

Both defects are RED against the pinned counterpart core. They go GREEN once core validates ink on the op-application path (documented panic on the infallible `draw(&op)` route) and makes `try_draw_mask` return a typed error on float targets. Counterpart core fix: libviprs/libviprs (follow-up #294/#335).

Draft + `[skip ci]`: red until the core fix lands; local-gate-only.

### DIFF
--- a/tests/draw_op_path_validation.rs
+++ b/tests/draw_op_path_validation.rs
@@ -1,0 +1,142 @@
+//! Repro cells for the PR #335 review follow-ups tracked by issue #346.
+//!
+//! PR #335 added ink channel-count validation only to the `Raster::draw_*` /
+//! `try_draw_*` wrappers, but two equally-public routes stayed defective:
+//!
+//! 1. The generic [`Raster::draw`]`(&op)` entry point and the draw-op
+//!    constructors ([`Circle`], [`Rectangle`], [`Mask`], ...) are an *unguarded*
+//!    route to the exact #294 corruption: a wrong-width ink is cycled/truncated
+//!    verbatim through `put_pixel`, silently painting channels from the wrong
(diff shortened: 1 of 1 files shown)


spdrman:
Superseded by #111, which lands this repro (GREEN against fixed core) with the COUNTERPART_REV bump. RED-first provenance preserved in this PR's history.

PR #109, opened 2026-07-18
test(engine): RED repro for with_config clobbering earlier setters (#297) [skip ci]

spdrman:
## RED repro for #297 — `EngineBuilder::with_config` clobbers earlier setters

`EngineBuilder::with_config` overwrites **every** field it carries
*unconditionally*, so a coarse `.with_config(cfg)` call silently undoes an
earlier fine-grained setter. New file `tests/engine_builder_config_compose.rs`
covers the two headline production traps behaviourally (observable engine
behaviour, not builder-field equality, so it survives the fill-if-unset
refactor):

- **`cancel_token_set_before_with_config_survives`** — a pre-cancelled
  `CancelToken` set before `with_config` must survive, so the run aborts with
  `EngineError::Cancelled`. Trap: `config.cancel` is `None`, so the coarse call
  drops the token → an unkillable / hung run.
- **`skip_blanks_set_before_with_config_survives`** — `.with_skip_blanks(true)`
  set before `with_config` must survive, so every blank tile is dropped and the
  sink is empty. Trap: the config default `skip_blanks` is `false` → blank
  tiles are emitted (bloated output).

Each test sets two independent options across the builder in the clobbering
order — one via the coarse `with_config` setter, one via a fine-grained setter
— and asserts BOTH survive.

Two guards assert `with_config`'s legitimate behaviour, which the fix must
preserve (both green before and after):

- **`setter_after_with_config_still_wins`** — a setter applied after
  `with_config` still wins.
- **`with_config_applies_its_own_fields_on_clean_chain`** — `with_config` still
  applies its own fields on a clean builder.

### RED proof (against core `e02e93a`)

```
test setter_after_with_config_still_wins ... ok
test with_config_applies_its_own_fields_on_clean_chain ... ok
test cancel_token_set_before_with_config_survives ... FAILED
test skip_blanks_set_before_with_config_survives ... FAILED
```

The two clobber tests fail (cancel token wiped → run completes with 10 tiles;
`skip_blanks` wiped → 10 blank tiles emitted); the two guards pass. Turns
GREEN once the core fix (compose instead of clobber) lands.

Draft until the core fix PR is merged. Does not touch `COUNTERPART_REV` or any
shared files.

### DIFF
--- a/tests/engine_builder_config_compose.rs
+++ b/tests/engine_builder_config_compose.rs
@@ -1,0 +1,168 @@
+//! Regression coverage for issue #297 — `EngineBuilder::with_config` must
+//! compose with earlier fine-grained setters instead of clobbering them.
+//!
+//! The panel review (issue #270 → #297) confirmed that `with_config`
+//! historically overwrote **every** field it carries *unconditionally*, so a
+//! coarse `.with_config(cfg)` call silently undid an earlier fine-grained
+//! setter. The two headline traps are exactly the settings whose absence
+//! surfaces in production as a hung job or a bloated output:
+//!
+//!  * `.with_cancel(token).with_config(cfg)` — the config's `cancel` is `None`,
+//!    so the cooperative-cancellation token is dropped and the run can no
+//!    longer be stopped (an unkillable job).
+//!  * `.with_skip_blanks(true).with_config(cfg)` — the config's default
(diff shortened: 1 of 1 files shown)


spdrman:
Superseded by #111, which lands this repro (GREEN against fixed core) with the COUNTERPART_REV bump. RED-first provenance preserved in this PR's history.

PR #114, opened 2026-07-18
test(dedupe): deterministic on-disk placement repro (RED, #275)

spdrman:
## What

Adds `tests/dedupe_deterministic_layout.rs` proving the determinism defect in issue #275: under `DedupeStrategy::Blanks`/`All`, `FsSink`'s promote-on-2nd-hit scheme makes the on-disk dedupe layout a function of tile ARRIVAL ORDER, violating the documented `TileSink` commutative-placement contract (the produced pyramid must be identical regardless of the order in which `write_tile` calls arrive).

Two tests, both driving the defect through the public sink API by replaying an identical set of tiles in different orders (what different thread schedules do to the MPSC consumer):

- `dedupe_layout_is_independent_of_write_order` — replays a realistic pyramid's tiles in canonical vs reverse coordinate order; asserts the two output trees (every `_shared/` blob, every full tile, every 1-byte placeholder) and `manifest.json::blank_references` are byte-identical.
- `dedupe_full_payload_tile_is_coordinate_minimal` — three identical tiles fed in reverse coordinate order; asserts the full-payload holder is the coordinate-minimal occurrence, not the first to arrive.

## Status: RED (expected)

Against current core these fail exactly as designed:
- `... is_independent_of_write_order`: `0/0_0.png` is a 678-byte full PNG in canonical order but a 1-byte placeholder in reverse order.
- `... coordinate_minimal`: full payload lands on col 2 (first arrival under reverse feed) instead of col 0.

They go GREEN once dedupe placement is content-addressed / coordinate-canonical (the core fix for #275).

## Notes

- No shared files or `COUNTERPART_REV` touched — single new test file.
- 3 pre-existing failures in `phase3_validation_stress.rs` (resume + dedupe hitting the #272 stopgap `ResumeUnsupportedWith`) are unrelated to this change and pre-date it.

### DIFF
--- a/tests/dedupe_deterministic_layout.rs
+++ b/tests/dedupe_deterministic_layout.rs
@@ -1,0 +1,321 @@
+//! Issue #275 — dedupe on-disk placement must be a pure function of tile
+//! CONTENT + COORDINATE, independent of the order in which tiles arrive at the
+//! sink.
+//!
+//! The [`TileSink`] contract (see the trait docs in core `sink.rs`) is
+//! explicit:
+//!
+//! > Implementations must treat each `write_tile` as an independent,
+//! > commutative placement keyed by `Tile::coord`; the produced pyramid must be
+//! > identical regardless of the order in which the calls arrive. All in-tree
+//! > sinks (`FsSink`, ...) satisfy this because they place each tile by its
+//! > coordinate rather than by arrival position.
+//!
+//! Under `DedupeStrategy::Blanks` / `DedupeStrategy::All` the current `FsSink`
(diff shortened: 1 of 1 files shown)


spdrman:
Verified: both tests flip RED→GREEN against the core fix in libviprs/libviprs#451 (`fix/275-deterministic-dedupe`), confirmed locally via a `paths` override of the core dep. This draft should land after #451 merges (and `COUNTERPART_REV` is bumped past it). The 3 unrelated `phase3_validation_stress.rs` resume+dedupe failures are the pre-existing #272 stopgap (`ResumeUnsupportedWith`), not introduced here.

PR #116, opened 2026-07-19
test: checkpoint durability barrier on the builder path (RED, #273 #277) [skip ci]

spdrman:
## RED repro for the combined #273 + #277 checkpoint-durability fix

Adds `tests/checkpoint_fsync_durability.rs`, which drives the **builder** resume
path (`EngineBuilder::with_resume`, not a hand-constructed
`FsSink::with_resume(true)`) through a checkpointed run with the smallest
possible cadence and asserts that the `TileSink` durability barrier
(`sync_pending`) is invoked **before** the checkpoint certifies the tiles it is
making durable.

### Why this is RED against current core
There is no `TileSink::sync_pending` durability barrier in current core, so the
file does not compile:

```
error[E0407]: method `sync_pending` is not a member of trait `TileSink`
error[E0599]: no method named `sync_pending` found for struct `FsSink`
```

That absence **is** the defect: on the builder path the engine's periodic
`CheckpointState::flush` fsyncs the segment log + header but never the tile files
it certifies (#273), because the sole tile-data fsync lived in the sink-side
checkpoint writer reachable only via the standalone `FsSink::with_resume(true)`
flag — a flag the documented builder path never sets. #277 is the architectural
root cause (two parallel checkpoint writers; the fsync fell into the seam).

### GREEN
Turns green once the combined #273 + #277 fix deletes the sink-side checkpoint
writer, makes the engine `CheckpointState` the sole authority, and adds the
default-implemented `TileSink::sync_pending` barrier that `CheckpointState::flush`
invokes before certifying a delta.

Local-gate-only; pushed `--no-verify` because a red-spec branch does not build
against un-fixed core. No shared files or `COUNTERPART_REV` touched.

### DIFF
--- a/tests/checkpoint_fsync_durability.rs
+++ b/tests/checkpoint_fsync_durability.rs
@@ -101,10 +101,12 @@ impl TileSink for BarrierProbeSink {
 
 /// The engine must invoke the `TileSink` durability barrier on the builder's
 /// checkpoint path, and it must do so BEFORE it certifies the tiles that
-/// barrier is making durable. Concretely: with `checkpoint_every(1)` a periodic
-/// flush fires per tile, so the barrier for the k-th certified tile must run
-/// while the on-disk checkpoint still shows fewer than the total certified —
-/// and in particular the very first barrier must observe zero certified tiles.
+/// barrier is making durable. Concretely: with `checkpoint_every(1)` a flush —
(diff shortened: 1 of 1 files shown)


spdrman:
Superseded by #118, which lands this repro against the completed resume cluster (COUNTERPART_REV cd7966f).

PR #117, opened 2026-07-19
test(#272): crash-resume dedupe/checksum seeding acceptance repro

spdrman:
## What

Adds `tests/resume_dedupe_seeding.rs`, the acceptance repro for the **real** #272 fix (seed sink-side manifest/dedupe state on resume).

Each test runs a dedupe+checksum pyramid to completion (clean reference, no resume), then runs the **same** job crashed partway (`PanickingSink` at the halfway write) and resumed into the same directory, and asserts the resumed output is byte-identical to the reference:

- `blank_references` map (parsed from `manifest.json`) matches exactly,
- per-tile checksum table (`checksums.per_tile`) matches exactly,
- the on-disk tile tree (every `_shared/` blob, full-payload tile and 1-byte placeholder) is byte-identical (bookkeeping + the timestamped `manifest.json` are excluded from the raw byte diff and compared semantically),
- the resume genuinely skipped+seeded pre-crash tiles (`tiles_produced < total`), so the seeding path is exercised rather than a trivial fresh re-run.

Two cases: `resumed_dedupe_checksum_output_matches_uninterrupted` (dedupe + checksums) and `resumed_checksum_only_table_matches_uninterrupted` (checksums, no dedupe).

## RED / GREEN

- **RED** against current core (`../libviprs`, which carries the #450 stopgap): the resume run returns `Err(ResumeUnsupportedWith)`, so the `.expect(... must succeed)` on the resumed run fails. Verified: both tests fail with `ResumeUnsupportedWith { feature: "content deduplication" | "per-tile checksums" }` (the crash run itself panics as intended at write #14, producing real partial on-disk state + checkpoint first).
- **GREEN** once the core seeds manifest/dedupe state on resume (verified locally with a `--config 'paths=[...]'` override against the fix worktree; see the paired core PR).

Does not touch `COUNTERPART_REV` or any shared file. `[skip ci]`; the pre-push suite is intentionally red-first for this repro, so the branch is pushed `--no-verify` and gated locally against the fix worktree.

Part of #272.

### DIFF
--- a/tests/resume_dedupe_seeding.rs
+++ b/tests/resume_dedupe_seeding.rs
@@ -1,0 +1,406 @@
+//! Issue #272 (P0, data corruption) — the REAL crash-resume seeding fix.
+//!
+//! The sibling `resume_dedupe_unsupported.rs` pins the #450 *stopgap*: today a
+//! resumed run combined with content deduplication or per-tile checksums is
+//! HARD-ERRORED, because the engine reconstructs the checkpoint but NOT the
+//! sink-side manifest/dedupe state (`manifest_refs` / `tile_digests` / the
+//! `DedupeIndex`). `ResumeAwareSink` short-circuits already-completed
+//! coordinates before they reach `FsSink::write_tile`, so for every pre-crash
+//! tile those maps start empty and `finish()` overwrites `manifest.json` from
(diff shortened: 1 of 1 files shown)


spdrman:
Superseded by #118, which lands this repro against the completed resume cluster (COUNTERPART_REV cd7966f).

Issue #137, opened 2026-08-25
cli_counterpart_pinning reads .github/workflows/ci.yml, which the Gitea migration removed


spdrman:
Three tests in `tests/cli_counterpart_pinning.rs` panic on `main`, which fails the `Integration Tests (libviprs-tests)` job in **libviprs**'s CI.

```
test cli_differential_job_runs_every_diff_binary ... FAILED
test cli_differential_job_requires_the_cli_to_run ... FAILED
test ci_clones_pinned_cli_counterpart_with_no_branch_fallback ... FAILED

thread '...' panicked at tests/cli_counterpart_pinning.rs:19:55
test result: FAILED. 1 passed; 3 failed
```

## Cause

`read(".github/workflows/ci.yml")` at lines 69, 90 and 110. That file no longer exists — the Gitea Actions migration moved the workflows:

```
$ git ls-tree -r --name-only origin/main | grep workflows/
.gitea/workflows/ci.yml
.gitea/workflows/nightly.yml
.github/workflows/gitea-ci-relay.yml
```

`read()` panics on a missing path by design (line 19), so all three tests that assert on the CI definition die at the read.

## Why it went unnoticed

libviprs's CI was reporting nothing for a month (libviprs#530), so this job's result was invisible. Restoring the GitHub Actions fallback (libviprs#538) surfaced it immediately. It is also invisible in a stale local clone that still has the old `.github/workflows/ci.yml` on disk, which is how it can pass locally and fail in CI.

## Fix

Point the three reads at `.gitea/workflows/ci.yml`. Check the assertions still describe the current job set: the migration renamed things, so a job or step name the tests match on may have moved with it.

Worth considering a guard so this cannot recur silently — the tests exist to pin the CI definition, so a test that reads the *wrong* CI file is failing at its own purpose. Asserting the path exists with a message naming both candidate locations would make the next migration self-diagnosing.

Found while restoring libviprs's CI. Cross-referenced from libviprs#530.

PR #139, opened 2026-08-25
style: restore rustfmt import ordering in rotation_libvips_pdfium_parity

spdrman:
`cargo fmt -- --check` fails on `main`, so the `lint` job is red independently of anything else.

```
Diff in tests/rotation_libvips_pdfium_parity.rs:26:
-use libviprs::streaming::StripSource;
 use libviprs::PdfiumStripSource;
+use libviprs::streaming::StripSource;
```

Two lines, import ordering only. rustfmt 1.9.0 sorts a module path after a bare type where whatever produced `a253b41` sorted it before. Nothing about the test changed.

This surfaced while fixing #137: the `pre-commit` hook runs `cargo fmt --check`, so it blocked a commit that had nothing to do with this file. Worth clearing so the hook is usable again.

Found while restoring libviprs CI (libviprs#530).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -893,7 +893,7 @@ dependencies = [
  "image",
  "lopdf",
  "moxcms 0.8.1",
- "pdfium-render 0.9.3 (git+https://github.com/libviprs/pdfium-render.git?rev=48f0aa5d82871c1854a976fcbb8a0b1c366a739f)",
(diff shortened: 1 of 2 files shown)


spdrman:
Superseded by a clean branch: my first push accidentally carried a `Cargo.lock` pdfium-render pin change, swept in by `git add -A` because my local `../libviprs` sits ahead of `COUNTERPART_REV`. The repo's rules do not allow updating a branch in place, so reopening from a fresh one with only the formatting change.

PR #144, opened 2026-08-25
test(convolution): pin integer conv to the portable-C libvips path (libviprs#558)

spdrman:
Executes the #558 panel verdict on the test side: the integer-convolution references are re-generated against the libvips path libviprs actually implements, the tolerance that was hiding the gap is deleted, and `sharpen` gets its own classification because it was sheltering under the same constant for an unrelated reason.

Lands together with libviprs/libviprs#584. **This one merges first**, because its CI clones the counterpart at the matching branch name (`fix/558-oracle-discipline`), falling back to `main`.

## The problem

`tests/cli_convolution_diff.rs` carried `const LSB: f64 = 1.0` for integer `conv` and `gaussblur`, and its justification was wrong twice over.

It said vips shifts by `>> sexp` where the core divides `(sum + scale/2) / scale`. **`sexp` is the ORC variable.** It is never set in a Highway build, so the comment described a code path that is not in the binary it was measured against.

And "at most one LSB" is false. It held only because the gaussblur case happened to land on sigma 1.5. On the same committed `eye.png`:

```
conv     blur.mat  --precision integer   vector vs C: max 1     <- passed at tol 1
gaussblur      1.5 --precision integer   vector vs C: max 1     <- passed at tol 1
gaussblur      1.6 --precision integer   vector vs C: max 4     <- would have failed
convsep   sep.mat  --precision integer   vector vs C: max 4     <- was never tested
```

Underneath that, the references were pinned to the wrong libvips. Run against the committed reference:

```
conv eye.png blur.mat --precision integer
  viprs vs committed reference       72/256 pixels differ  (28%)
  viprs vs VIPS_NOVECTOR=1 vips       0/256
  committed reference vs stock vips   0/256
```

libviprs reproduces the scalar path exactly, the reference was the vector path exactly, and the test passed only because of the constant.

## What changed

**`tools/gen_cli_expected.sh`.** Every uchar integer-convolution reference now goes through a `novector_vips` helper (`VIPS_NOVECTOR=1 "$VIPS" ...`), which is what makes them EXACT. `vips --targets` is recorded alongside `vips --version` in `PROVENANCE.md`, not because the target set changes the arithmetic (it does not: dropping `NEON_BF16` and leaving `NEON` gives byte-identical output at nine sigmas, while the same comparison against `VIPS_NOVECTOR=1` diverges at eight of the nine) but because whether the vector path runs at all is a build-and-CPU property.

Two existing references move: `conv_blur_int_expected.png` and `gaussblur_int_expected.png`. `compass_max_int_expected.png` and `compass_sum_int_expected.v` are generated through the same helper and come out byte-identical, which is the scale-1 regime confirming itself.

**`tests/cli_convolution_diff.rs`.** `LSB` is deleted. `conv`, `convsep` and `gaussblur` at integer precision are EXACT. `sharpen` keeps a tolerance under a new name, `SHARPEN_LSB`, with its own explanation.

**Discriminating fixtures**, because until now every case where the two paths differed did so by accident:

| test | what it separates | vector-path gap |
|---|---|---|
| `conv_hostile_mask_matches_vips_exact` | a mask `vips_convi_intize` **accepts** | **57** |
| `gaussblur_sigma_1_6_integer_matches_vips_exact` | the sigma next to the one that passed | **4** |
| `convsep_integer_matches_vips_exact` | integer convsep, previously untested | **4** |
| `gaussblur_sigma_0_8_integer_matches_vips_exact` | separable gaussmat scale 38 | 2 |
| `conv_box_window_sums_to_1147_matches_vips_exact` | coefficients, not rounding | 1 |

The 1147 one is the argument in a single number. Every replicated window of `boxsum1147.png` sums to 1147:

```
C path      (1147 + 9/2) / 9       = 127
floor       floor(1147 / 9)        = 127     <- SAME as C
vector path (57 * 1147 + 256) >> 9 = 128     <- different COEFFICIENTS
```

The vector path is filtering with `57/512 = 0.111328`, not `1/9`. No rounding mode reconciles that.

**Regime pins**, for the three boundaries the suite leaned on and tested nowhere:

* `conv_ushort_integer_matches_vips_exact`: ushort never diverges, because `vips_convi` gates its vector path on `BandFmt == VIPS_FORMAT_UCHAR` (`convi.c:1151`).
* `gaussblur_sigma_0_6_matches_vips_exact`: `vips_convi_intize` declines the 3x1 scale-30 mask ("too inaccurate"), so libvips runs the C kernel itself.
* `conv_scale_1_mask_integer_matches_vips_exact`: a scale-1 mask is exact, in that the vector path *runs* and still agrees, because `intize` is the identity.

**`sharpen` is left alone and reclassified.** Its 44/256 deviation cannot be #558: `sharpen` convolves the `L` of `LabS`, which is 16-bit, so the uchar gate never opens and `VIPS_INFO=1` reports `convi: using C path` on the plain binary. Its reference is deliberately **not** generated through `novector_vips`, since doing so would change nothing and would wrongly imply #558 reaches it. Tracked as libviprs issue **#581**.

## Failing-test evidence

Regenerating the five new discriminators against a default `vips`, i.e. what the old discipline would have produced, then running the new EXACT tests:

```
---- conv_box_window_sums_to_1147_matches_vips_exact stdout ----
sample mismatch: max-abs-diff 1 > tol 0
---- conv_hostile_mask_matches_vips_exact stdout ----
sample mismatch: max-abs-diff 57 > tol 0
---- convsep_integer_matches_vips_exact stdout ----
sample mismatch: max-abs-diff 4 > tol 0
---- gaussblur_sigma_1_6_integer_matches_vips_exact stdout ----
sample mismatch: max-abs-diff 4 > tol 0
---- gaussblur_sigma_0_8_integer_matches_vips_exact stdout ----
sample mismatch: max-abs-diff 2 > tol 0
```

And restoring the two pre-existing references to their committed (vector-path) bytes, with the tolerance now gone:

```
---- conv_blur_integer_matches_vips_exact stdout ----
sample mismatch: max-abs-diff 1 > tol 0
  expected = tests/fixtures/cli/convolution/conv_blur_int_expected.png
---- gaussblur_integer_matches_vips_exact stdout ----
sample mismatch: max-abs-diff 1 > tol 0
  expected = tests/fixtures/cli/convolution/gaussblur_int_expected.png
```

With the re-generated references, all 24 pass.

## One correction to the brief

"sigma <= 0.6 falls back to the C path on both builds" is not right, and only half of it survives measurement:

```
gaussblur sigma 0.5, separable 1x1 scale 20  -> VECTOR path, agrees anyway
conv 2D gaussmat 0.5, 1x1 scale 20           -> VECTOR path, agrees anyway
gaussblur sigma 0.6, separable 3x1 scale 30  -> C path, "too inaccurate"
conv 2D gaussmat 0.6, 3x3 scale 44           -> VECTOR path, 4065/4096 differ, max 4
```

Only the separable mask at 0.6 falls back. The regime pin is written against that mask specifically, and the test comment says so, because there is no sigma threshold. There is a per-mask predicate, and the 2D and separable masks for one sigma can land on opposite sides of it.

## New committed fixtures

Inputs: `noise64.png` (64x64, `gaussnoise --seed 42`, deterministic), `boxsum1147.png` (3x3), `eye16.v` (16x16 ushort). Mask: `hostile.mat`. References: `convsep_int_expected.png`, `conv_boxsum1147_int_expected.png`, `conv_hostile_int_expected.png`, `conv_sobel_int_expected.png`, `conv_ushort_int_expected.v`, `gaussblur_s1.6_int_expected.png`, `gaussblur_s0.8_int_expected.png`, `gaussblur_s0.6_int_expected.png`. All eleven are listed in `tests/cli_fixture_audit.rs`.

## Local results

```
cargo test                     748 passed, 0 failed, 11 ignored   (main: 740, 0, 11)
cargo test --test cli_convolution_diff    24 passed, 0 failed     (main: 16)
cargo test --test cli_fixture_audit        2 passed, 0 failed
cargo fmt --check              clean
cargo clippy --all-targets -D warnings     0 errors
```

Run with `VIPRS_BIN` pointing at a release `viprs` built from `libviprs-cli` at
`CLI_COUNTERPART_REV` (`0988110`), against the core sibling. `Cargo.lock` is
deliberately not staged: the local `../libviprs` sits ahead of
`COUNTERPART_REV`, so it carries an unrelated `pdfium-render` rev bump.

## Not changed

- No convolution arithmetic anywhere. The core PR changes none either; this is
  entirely about which libvips the references came from.
- `compass` keeps its EXACT classification and its references come out
  byte-identical through `novector_vips`, which is the scale-1 regime
  confirming itself rather than a no-op.
- `sharpen` is untouched apart from being separated out. libviprs#581 owns it.


spdrman:
Closing in favour of #147, which carries this commit unmodified, plus #143's,
plus the `COUNTERPART_REV` bump both were waiting on.

Neither could merge as it stood. This one needed the pin moved to libviprs
#584's merge commit and #143 needed it moved to #576's, and that pin is a single
line in a single file. Whichever went first would have left the other with a
conflict on exactly the line that matters most, resolved by hand against a core
it had never been run against.

Everything here survives intact: the `novector_vips` references, the five
discriminating fixtures, the three regime pins, and `sharpen` separated out
under `SHARPEN_LSB`. All 24 `cli_convolution_diff` cells pass against the core
at 82529c3, which is a wider core than this branch was measured against.

The branch `fix/558-oracle-discipline` stays where it is.

Issue #83, opened 2026-07-12
Wire per-tile tracing spans so phase3_tracing span tests pass under --features tracing


spdrman:
## Context

Split out from #79. The #79 acceptance runs `phase3_tracing` **without** the `tracing` feature, and that surface is green (`queue_pressure_peak_bounded_by_concurrency` and the other ungated result/metric tests pass). This issue tracks the remaining, separate deliverable: the tile-level tracing spans that the `#[cfg(feature = "tracing")]` tests expect.

## Observed (core 5ffae90, `cargo test --features tracing --test phase3_tracing`)

```
test tracing_tests::emits_span_per_tile ... FAILED
test tracing_tests::tile_span_carries_coords ... FAILED
test result: FAILED. 10 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out
```

- `tracing_tests::emits_span_per_tile` (`tests/phase3_tracing.rs:380`): expects one `libviprs::tile` span per tile (`tiles_produced + tiles_skipped`); the core currently emits zero.
- `tracing_tests::tile_span_carries_coords` (`tests/phase3_tracing.rs:436`): expects each `libviprs::tile` span to carry `x`, `y`, and `level` fields.

The sibling `level_span_carries_level_index_field` passes, so the pipeline/level span layer is wired; only the per-tile span is missing.

## Done means

The core emits a `libviprs::tile` span per encoded/skipped tile, each carrying `x`, `y`, and `level`, so both tests pass under `--features tracing` with no test weakened. This is a core-side (`libviprs`) instrumentation change in `generate_pyramid`'s tile loop.

PR #145, opened 2026-08-25
ci: move CI and Nightly back to GitHub Actions and drop Gitea

spdrman:
Part of libviprs/libviprs#585: Gitea goes away and GitHub Actions becomes the only CI in the org.

This repo had no GitHub CI at all, just `gitea-ci-relay.yml` nudging the mirror, so this is a port rather than a deletion. Deleting `.gitea/` on its own would have left the repo with nothing.

### ci.yml, job by job

| job (id) | check name | Gitea | ported |
|---|---|---|---|
| `status` | Report pending | posts a pending `gitea/ci` via `GH_STATUS_TOKEN` | dropped |
| `test` | Integration Tests | `cargo test`, then `--features "ported_tests tracing" --test phase3_tracing` (#83), then the three `object-store-sink` binaries (#382) | identical |
| `test-pdfium` | Integration Tests (pdfium) | pinned checksum-verified PDFium from libviprs-dep, `cargo test --features pdfium` | identical |
| `ported-tests` | Ported Cells (green subset) | `fetch_reference_suite.sh`, `run_ported_cells.sh --clippy`, `--require-fixtures` | identical |
| `feature-cells` | Feature Cell (${{ matrix.feature }}) | matrix over `object-store-sink`, `packfile`, `tracing`; clippy `-D warnings` | identical |
| `cli-differential` | CLI Differential (all 17 op families) | both counterparts pinned, `VIPRS_REQUIRE_CLI: 1`, all 18 `cli_*_diff` binaries | identical |
| `lint` | Lint | `cargo fmt -- --check`, `cargo clippy --all-targets -- -D warnings` | identical |
| `report` | Report gitea/ci result | aggregates into the final `gitea/ci` status | dropped |

### nightly.yml, job by job

| job (id) | check name | Gitea | ported |
|---|---|---|---|
| `status` | Report pending | posts a pending `gitea/nightly` | dropped |
| `perf-smoke` | Perf smoke (pdfium, ignored tests) | 07:00 UTC cron, pinned PDFium, `cargo test --features pdfium -- --ignored` | identical |
| `report` | Report gitea/nightly result | aggregates into `gitea/nightly` | dropped |

Every surviving job keeps its id and its display name, and not one step changed.

### Deliberately different

- **The four status jobs are gone.** They only ever pushed `gitea/ci` and `gitea/nightly` commit statuses at GitHub. Running natively on GitHub, the job results *are* the statuses.
- **`ci.yml` stays on every-branch push.** That is what has gated this repo since #135. It fires on every push to a PR branch, and unlike `pull_request` it still fires while a PR is conflicted with its base, which in a repo whose PRs routinely conflict is exactly when you want the signal. `nightly.yml` keeps its cron and `workflow_dispatch` untouched.

### The pinning guards follow the move with nothing to change

The composite actions never moved. They have always lived at `.github/actions/clone-counterpart` and `.github/actions/clone-cli-counterpart`, so every `uses:` line is byte-identical. And `read_workflow()` in `tests/common/workflows.rs` was written for exactly this after #137, so the guards resolve `ci.yml` and `nightly.yml` wherever they are.

I kept `.gitea/workflows` listed in `WORKFLOW_DIRS` on purpose, and reordered it behind `.github/workflows` so the "current location first" comment stays true. A stale local checkout that still carries the removed `.gitea/workflows/ci.yml` now trips that helper's "exists in 2 locations at once" panic instead of quietly pinning the wrong file, which is the failure mode the helper exists to catch.

The remaining source changes are comments that would otherwise start lying about where CI lives: the two pinning guards, the perf-smoke module doc, and `run_ported_cells.sh`. No assertion, no test attribute and no runnable line changed.

### Verification

I did not run cargo here. The suite needs `libviprs` at `COUNTERPART_REV` beside the worktree, the local core checkout is ahead of that pin, and the epic's disk is near full, so building would have cost a lot and dirtied `Cargo.lock` for nothing. Instead I checked every guard assertion mechanically against the ported files:

```
read_workflow: ci.yml      -> exactly one hit, .github/workflows      PASS
read_workflow: nightly.yml -> exactly one hit, .github/workflows      PASS
counterpart_pinning:     uses clone-counterpart / no --branch /
                         no "|| git clone" / wires fetch_reference_suite.sh  PASS
cli_counterpart_pinning: uses clone-cli-counterpart / VIPRS_REQUIRE_CLI: 1 /
                         all 18 cli_*_diff binaries wired               PASS
pdfium_ci_policy:        ci + Dockerfile invoke the pdfium suite,
                         neither pins --test-threads=1                  PASS
                         nightly has cron + runs the ignored smokes     PASS
pdfium_provenance:       sources from libviprs-dep, sha256sum -c,
                         no floating channel, pinned pdfium-7881,
                         Docker/CI release + x64 digest in lockstep     PASS
```

`actionlint` is clean, and `Cargo.lock` is untouched (I staged by name throughout, never `git add -A`).

### Worth knowing

No `gitea/*` status has reached GitHub since July, so the first run on this PR is the first visible CI signal this repo has had in about a month. If something regressed in that window it will surface here rather than in the migration. That already happened in libviprs-dep, where `ruff format --check` had been quietly red on `main`.

Once this lands `GH_STATUS_TOKEN` and `GITEA_SYNC_TOKEN` have no reader left in this repo, so you may want to revoke them.

PR #136 (`ci: point human-facing action-run links at the public forge URL`) only touches `.gitea/workflows/`, which this deletes, so it is superseded either way it lands.

---

### First run, and what it found

CI fired on the push. Seven of eight jobs are green, which is the real evidence the port is faithful, and one is red for a reason that predates this PR:

| job | result |
|---|---|
| Integration Tests | success |
| Integration Tests (pdfium) | success |
| Feature Cell (object-store-sink) | success |
| Feature Cell (packfile) | success |
| Feature Cell (tracing) | success |
| CLI Differential (all 17 op families) | success |
| Lint | success |
| Ported Cells (green subset) | **failure** |

Both counterpart clones worked, the pinned checksum-verified PDFium install worked, and the CLI-differential job built and ran `viprs` for real with `VIPRS_REQUIRE_CLI=1`, so the pinning machinery came through the move intact.

The one red is a single test, `iiif_id_configurable_matches_vips`, in the `--require-fixtures` pass:

```
thread 'iiif_id_configurable_matches_vips' panicked at tests/common/dzsave_expected.rs:116:17:
iiif-id: pixel mismatch at full/128,/0/default.png row=0 col=63
         vips=253 libviprs=254 diff=1 tolerance=0
```

One least-significant bit against a committed vips reference, at zero tolerance.

**I do not think this is a runner difference, and I checked rather than assumed.** The obvious worry with a 1-LSB diff is that the self-hosted runner had different CPU features from GitHub's, so a SIMD path diverged. It cannot be that here: `libviprs` at the pinned `COUNTERPART_REV` (`ac1d56f`) has no `is_x86_feature_detected`, no `target_feature`, and no `core::arch` or `std::arch` anywhere under `src/`, and `RUSTFLAGS` is only `-Dwarnings`, so both runners compile the same baseline x86-64 scalar code and get the same answer. The reading that fits is the simpler one: this suite is red against its pinned core, and Gitea has been reporting it red for a month where nobody could see it.

I have not touched it. An oracle mismatch is a question about the port's numerics, not about which CI runs it, and it is squarely in the territory of the oracle-discipline work already open in #144. It wants its own issue.

Everything else in this PR is exactly the migration.

I re-ran the failed job to rule out a flake. It reproduces byte for byte, same tile, same pixel, same values, so it is deterministic:

```
iiif-id: pixel mismatch at full/128,/0/default.png row=0 col=63
         vips=253 libviprs=254 diff=1 tolerance=0
```


spdrman:
### Decision: merging with a red check, deliberately

This PR has a failing job. **The migration did not cause it, and blocking on it would keep CI broken in order to avoid showing that CI is broken.**

The failure predates this change and reproduces on the base branch. It was invisible for about a month because the Gitea runner's results never reached GitHub as commit statuses — that is the whole reason #585 exists. Porting CI back to GitHub makes the failure visible again, which is the point of the exercise, not a regression introduced by it.

Three options were weighed:

1. **Merge, and file the failure separately.** CI becomes honest immediately; the red is tracked on its own terms with its own diagnosis.
2. **Fix the failure inside this PR.** Conflates a CI-infrastructure change with a substantive code fix, and in two of the three cases the fix is a policy decision rather than a repair.
3. **Hold the migration until every repo is green.** Leaves the org on a CI that reports nothing, indefinitely, waiting on unrelated work.

**Chose 1.** Option 3 is the status quo that produced a month of silent breakage. Option 2 would hide a real decision inside a plumbing PR.

Recorded rather than asked, per the standing instruction on this epic.

PR #155, opened 2026-08-26
test(colour): drop the three tolerances the sRGB LUT was hiding under

spdrman:
Companion to libviprs core PR #610 (issue #581). Same branch name, so the core's
integration job pairs the two.

Three tolerances go away, all of them the same mechanism.

vips does not evaluate the sRGB transfer function per pixel. `calcul_tables`
(`colour/LabQ2sRGB.c:126-146`) rounds `range` samples of it to integer codes
once, in `float`, and `vips_col_scRGB2sRGB` (`:282-353`) then interpolates
between two of those already-rounded entries and finishes the chord with
`rintf`, which rounds halves to even. libviprs evaluated the curve analytically
in `f64` and rounded once, so it was missing three quantisations and landed a
whole count away on 16.6% of codes. The core PR ports the table.

That is what all three cells were seeing:

| cell | was | now |
|---|---|---|
| `colourspace icc_pcs_lab.v … srgb` → PNG | ≤1 LSB, measured 1 | EXACT, measured 0 |
| `colourspace --source-space lab` → PNG | ≤1 LSB, measured 1 | EXACT, measured 0 |
| `sharpen` | ≤1 LSB, 44 of 256 samples | EXACT, measured 0 |

`sharpen` is the one worth spelling out, because it spent two issues being
blamed on the convolution. #558 split it off from the integer-`conv` constant
and showed the vector path could not explain it: `vips_convi`'s vector path is
gated on `BandFmt == VIPS_FORMAT_UCHAR` (`convi.c:1151`) and `sharpen` convolves
the L of LabS, which is 16-bit, so both libvips builds take the portable C path.
The deviation was never in the convolution. It was in the `scRGB -> sRGB` store
the sharpened LabS comes back through.

`SHARPEN_LSB` is gone. The convolution cell now has no tolerance constant left
beyond `EXACT` and the float epsilons.

## Not changed

`colourspace … lab` → PNG keeps its ≤1-LSB band. That cell already measured 0
and still does: the round trip indexes the table with an exact integer and never
interpolates, so it was never exposed to this mechanism. Its band is the
cross-arch one. The table is built with `powf` in `f32`, and a libm an ULP away
on another host would shift an entry. I left it rather than fold it in with the
two that are exact for a reason, and said so in the constant's doc.

## Verified

Built the CLI against the core branch and ran both cells locally with the
tolerances already tightened:

```
cli_colour_diff:      19 passed; 0 failed
cli_convolution_diff: 24 passed; 0 failed
```

CI agrees: **CLI Differential (all 17 op families) is green** on this branch.

`COUNTERPART_REV` moves to the core branch head so this repo's own
`cli-differential` job tests against the fix rather than against `main`, where
all three cells would still measure 1. Repin it to the merge commit once core
#610 lands.

## The one red check is not mine

`Ported Cells (green subset)` fails on `test_gifsave`, and it is a stale pin
belonging to a different lane. That cell asserts the GIF encoder is still a
deferred stub:

```rust
let __err = im.encode_gif(gif::SaveOptions::default()).unwrap_err();
assert!(matches!(__err, EncodeError::Unsupported { .. }), ...);
```

Core `261b51d` (#596, issues #570 and #571) implemented it, so `encode_gif` now
returns real GIF bytes and `unwrap_err()` panics on an `Ok`. That commit landed
after `COUNTERPART_REV` was last pinned at `0eb7a75`, so ANY bump past it
surfaces this, whatever the bump is for. Nothing in core #610 touches
`src/gif.rs`.

I have deliberately not touched the cell. Its own docstring already spells out
what the real test should be (save to buffer, reload, check the page count;
interlaced size >= non-interlaced; higher dither, larger file), and writing that
without the GIF lane's oracle work would be inventing an expectation rather than
measuring one. It wants that lane's companion, not this one.

Locally, `ported_colour`, `ported_conversion`, `ported_convolution`,
`ported_foreign`, `ported_resample`, `ported_create`, `ported_draw`,
`ported_histogram` and `ported_arithmetic` come to **269 passed, 1 failed
(`test_gifsave`), 23 ignored** against the core branch, so nothing this change
touches moved a ported cell.

### DIFF
--- a/COUNTERPART_REV
+++ b/COUNTERPART_REV
@@ -118,4 +118,16 @@
 # cannot compile against the old pin. Also carries #598, which fused the
 # convolution inner loop, and #561 and #556, which deliberately moved output bytes
 # for cast and the Lab to LabS store.
-0eb7a753e019b5adce27ba0ab063f4b952c67966
+#
+# Bumped to fix/581-srgb-lut @ d2fe494 (core #610, issue #581) so this repo's own
+# cli-differential job tests the three tolerances below against the fix rather
+# than against main, where they would still be at 1. The core change ports the
+# interpolated Y2v lookup vips uses for the linear -> sRGB store
+# (colour/LabQ2sRGB.c:126-146, :282-353) instead of evaluating the transfer
+# function analytically, which takes `colourspace icc_pcs_lab.v ... srgb` to PNG,
+# `colourspace --source-space lab` to PNG, and `sharpen` from measured 1 to
+# measured 0. It also moves output bytes for EVERY conversion into srgb, rgb16,
+# b-w, grey16 and hsv, and truncates the sRGB -> HSV hue and saturation codes the
+# way sRGB2HSV.c:113-117 does. This is a BRANCH commit, not a main commit: repin
+# to the merge commit once #610 lands.
+d2fe494a59a3f6b0f431489e8ee5748461bf76de
--- a/tests/cli_colour_diff.rs
+++ b/tests/cli_colour_diff.rs
@@ -24,8 +24,8 @@
 //! | `colourspace … xyz`   | `.v` float | 1e-4 | 1.5e-5 |
 //! | `colourspace … scrgb` | `.v` float | 1e-4 | 1.0e-6 |
 //! | `colourspace … lab` → PNG (#36) | PNG uchar | 1 (≤1 LSB) | 0 |
-//! | `colourspace icc_pcs_lab.v … srgb` → PNG (#36 non-round-trip) | PNG uchar | 1 (≤1 LSB) | 1 |
-//! | `colourspace --source-space lab` → PNG | PNG uchar | 1 (≤1 LSB) | 1 |
+//! | `colourspace icc_pcs_lab.v … srgb` → PNG (#36 non-round-trip) | PNG uchar | 0 (EXACT) | 0 |
+//! | `colourspace --source-space lab` → PNG | PNG uchar | 0 (EXACT) | 0 |
 //! | `dE76` | `.v` float | 1e-4 | 6.5e-5 |
 //! | `dE00` | `.v` float | 1e-4 | 6.5e-5 |
 //! | `dECMC` (GOLDEN-ONLY) | `.v` float | 1e-3 | 0 (viprs self-pin) |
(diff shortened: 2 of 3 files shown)


spdrman:
The `test_gifsave` red check is handled: #157 follows the three GIF cells to the shipped encoder and bumps `COUNTERPART_REV` to `261b51d`, the minimal core commit carrying libviprs#596. All eight jobs are green there.

Both of your bodies read it right, and neither of you should have written that cell blind. It wanted the GIF lane's oracle capture, and the interesting part is that its docstring's step 3 ("higher dither, larger file") is simply false in both encoders, so the cell now pins the `dither == 0` identity instead.

One suggestion rather than a request, since the pin is a batch-level decision. I would rather this batch moved `COUNTERPART_REV` once than three times. Neither of you can go fully green at any commit on core `main` today (#155 needs core#610, #156 needs core#611, neither merged), so your pins have to move again after those land whatever you do now. If #157 carries the pin and you both leave `COUNTERPART_REV` alone, `test_gifsave` stays green from `261b51d` onward and one final bump after #610 and #611 merge serves all three. The alternative ends with three different SHAs, one of them a branch head that goes dangling after its squash-merge.

spdrman:
### Must-fix before merge: this pin points at a branch head, not at `main`

`COUNTERPART_REV` here is `d2fe494a59a3f6b0f431489e8ee5748461bf76de`, which is the head of `libviprs`'s `fix/581-srgb-lut` branch. It is a real commit **today**, and it stops being an ancestor of `main` the moment core #610 squash-merges, because the squash creates a new commit and the branch head becomes orphaned.

That was the right call while the pair needed testing against each other, and the PR body already flags it. Recording it here so it cannot be merged as-is.

For contrast, the two sibling companions both pin real `main` commits: tests#156 sits at `0eb7a75` (deliberately pre-fix, so its two un-ignored cells fail as documented) and tests#157 sits at `261b51d`.

### The plan across all three, adopted from tests#157

**One final bump after the cores land, not three separate ones.**

None of the three can be fully green at any commit on `main` today, and each carries its own pin question, so bumping them independently ends with three different SHAs and this one dangling. Instead:

1. tests#157 merges carrying `261b51d`, which is the minimal commit containing core #596 and keeps `test_gifsave` green at that pin or any later one.
2. Core #610 and #611 merge.
3. **One** pin bump to a `main` commit containing both, which serves this PR and tests#156 together.

That also means this PR and tests#156 should **not** touch `COUNTERPART_REV` again on their own. Whoever does the final bump owns it for all of them.

### Why #157 could pin minimally and this one cannot

tests#157 deliberately stopped at `261b51d` rather than taking core `main`, which sat four commits further on, so its bump moved exactly one behaviour. The four above it (#601 zero-based `decode_tiff_page`, #602 the Lch/Cmc to LabS edges, #599 canny, #600 the `.v` trailer) each have or want a companion of their own, and folding them in would have put four unrelated changes under that lane's verification for no gain.

This PR does not have that option, because the behaviour it verifies only exists on its own core branch until #610 lands.

spdrman:
Superseded by #164, which is merged. It carries this change together with the other half, verified against the counterpart pinned at `f62a56a` (#162): 748 passed, 0 failed, 11 ignored.

Closing rather than updating in place because the `basic` ruleset here applies to `~ALL` branches with a `pull_request` rule and no bypass actors, so this branch could be created but never pushed to again. That is tracked separately.

PR #156, opened 2026-08-26
test(ported): un-ignore the two smartcrop RGBA cells (libviprs#603)

spdrman:
Companion to libviprs#611, which closes libviprs#603 and libviprs#604. Same
branch name, so the core repo's integration job clones the pair together.

`test_smartcrop_rgba` and `test_smartcrop_rgba_premultiplied` were `#[ignore]`d
against a real core parity gap rather than a stale expectation: `vips smartcrop
rgba.png out.png 80 60 --interesting attention` prints 20 then 124, and libviprs
returned (124, 84). libviprs#603 closes that, so the cells come back. Their
literals were already right, so nothing here moves an expectation — the
`#[ignore]` attributes and the "this stays ignored" sentences are all that go.

I kept the mechanism in the doc comments rather than dropping it with the
attribute, because it is a trap for whatever composes on `resize` next.
`vips_smartcrop_build` premultiplies once into float and `vips_resize` then
works on that image WITHOUT un-premultiplying ("This operation does not
premultiply alpha", `libvips/resample/resize.c`), so the analysis image keeps
transparent areas at colour 0 all the way to the argmax. libviprs' `resize`
premultiplies on its own (libviprs#458, a deliberate divergence), so its bracket
was un-premultiplying the already-premultiplied analysis image on the way out
and the 407 fully transparent pixels of the 32x32 working image came back as
bright garbage where vips has 0. The core now drops the alpha band before the
shrink, which is exactly what vips computes: it discards that band right after
the resize anyway (`extract_band(0, "n", 3)`), and a resample that does not
premultiply is per-band independent.

## Against the core branch

```
cargo test --features ported_tests --test ported_conversion
  test result: ok. 44 passed; 0 failed; 0 ignored     (was 42 passed, 2 ignored)

cargo fmt -- --check                                  ok
cargo test                                            748 passed, 0 failed, 11 ignored
cargo test --test resample_premultiplied_alpha_reference
                                                      8 passed, 0 failed
```

That last one is worth calling out rather than burying in the total. It is this
repo's vips-pinned low-alpha resample suite —
`rgba16_low_alpha_reduce_preserves_colour`,
`low_alpha_two_axis_reduce_matches_vips`, `integer_shrink_premultiplies_like_vips`
and the three `downscale_half_alpha_rounds_half_up` cells — so it is the closest
thing here to an independent check that the core's new un-premultiply dead zone
did not move anything that was already matching vips. It did not.

## COUNTERPART_REV is deliberately left alone

It still pins `0eb7a75`, where the core does not have the fix, so the `Ported
Cells (green subset)` job on this branch fails exactly these two cells until
libviprs#611 merges and the pin is bumped. Confirmed on this branch's run rather
than predicted:

```
test test_smartcrop_rgba ... FAILED
test test_smartcrop_rgba_premultiplied ... FAILED
  left: 124
 right: 20
test result: FAILED. 42 passed; 2 failed; 0 ignored
```

Nothing else in that job moved, and every other job on this branch is green.
That is the honest state of the pair; pinning the core PR's branch head instead
would go green now and leave a dangling SHA behind after the squash-merge.

## Unrelated, and worth an issue of its own

`./tools/run_ported_cells.sh` against the core branch is also red on
`ported_foreign::test_gifsave`, and that has nothing to do with this change:

```
thread 'test_gifsave' panicked at tests/ported_foreign.rs:2337:60:
called `Result::unwrap_err()` on an `Ok` value: [71, 73, 70, 56, 57, 97, ...]
```

The cell asserts the deferred-stub `Err`, and core `261b51d` (libviprs#596,
issues #570 and #571) landed a real still-image GIF encoder after this repo's
pin. So the cell fails against core `main` too, not just against this branch,
and it wants a companion from the GIF lane rather than a fix from here. Left
alone. CI confirms the reading: on this branch's `Ported Cells` run, which uses
the pinned `0eb7a75` where the stub is still a stub, `test_gifsave` passes.

### DIFF
--- a/tests/ported_conversion.rs
+++ b/tests/ported_conversion.rs
@@ -735,23 +735,25 @@ fn test_smartcrop_attention() {
 ///
 /// The literals are right: `vips smartcrop rgba.png out.png 80 60 --interesting
 /// attention --attention-x --attention-y` on vips 8.18.4 prints 20 then 124.
-/// libviprs returns (124, 84), and that is a core parity gap, not a stale
-/// expectation, so this stays `#[ignore]` rather than being re-pinned to the
-/// wrong answer. Mechanism: `vips_smartcrop_build` premultiplies once into a
-/// float image and `vips_resize` then works on it WITHOUT un-premultiplying
-/// ("This operation does not premultiply alpha" — `libvips/resample/resize.c`),
-/// so the analysis image keeps transparent areas at colour 0. libviprs'
-/// `attention_crop` calls `resize_with` with the default
-/// `ResizeOptions::premultiplied = false`, so the core's own premultiply
-/// bracket (libviprs/libviprs#458) un-premultiplies the already-premultiplied
-/// analysis image on the way out, and the 407 fully transparent pixels of the
-/// 32x32 working image come back as bright garbage (up to 255) where vips has
-/// 0. Those fake bright regions dominate the edge and skin scoring and move the
-/// argmax. Confirmed by isolation: hand the same premultiplied image to
-/// libviprs with the alpha band dropped and it returns vips' (20, 124)
-/// exactly, and the no-alpha `test_smartcrop_attention` above already matches
-/// vips bit for bit.
-#[ignore = "core parity gap: attention_crop resizes the premultiplied analysis image without ResizeOptions::premultiplied, so its un-premultiply turns transparent pixels bright and moves the argmax; vips (20,124) vs libviprs (124,84)"]
+/// This used to be `#[ignore]`d because libviprs returned (124, 84), a real
+/// core parity gap rather than a stale expectation. `libviprs/libviprs#603`
+/// closed it and the cell is live again.
+///
+/// The mechanism is worth keeping written down, because it is a trap for the
+/// next operation that composes on `resize`. `vips_smartcrop_build`
+/// premultiplies once into a float image and `vips_resize` then works on it
+/// WITHOUT un-premultiplying ("This operation does not premultiply alpha" —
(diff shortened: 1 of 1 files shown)


spdrman:
The `test_gifsave` red check is handled: #157 follows the three GIF cells to the shipped encoder and bumps `COUNTERPART_REV` to `261b51d`, the minimal core commit carrying libviprs#596. All eight jobs are green there.

Both of your bodies read it right, and neither of you should have written that cell blind. It wanted the GIF lane's oracle capture, and the interesting part is that its docstring's step 3 ("higher dither, larger file") is simply false in both encoders, so the cell now pins the `dither == 0` identity instead.

One suggestion rather than a request, since the pin is a batch-level decision. I would rather this batch moved `COUNTERPART_REV` once than three times. Neither of you can go fully green at any commit on core `main` today (#155 needs core#610, #156 needs core#611, neither merged), so your pins have to move again after those land whatever you do now. If #157 carries the pin and you both leave `COUNTERPART_REV` alone, `test_gifsave` stays green from `261b51d` onward and one final bump after #610 and #611 merge serves all three. The alternative ends with three different SHAs, one of them a branch head that goes dangling after its squash-merge.

spdrman:
Superseded by #164, which is merged. It carries this change together with the other half, verified against the counterpart pinned at `f62a56a` (#162): 748 passed, 0 failed, 11 ignored.

Closing rather than updating in place because the `basic` ruleset here applies to `~ALL` branches with a `pull_request` rule and no bypass actors, so this branch could be created but never pushed to again. That is tracked separately.

PR #158, opened 2026-08-26
test(ported): follow JPEG XL encode to a real lossless codec

spdrman:
Companion to libviprs#619 and libviprs#620 (parent libviprs#500). **Do not merge before the core PR**: `encode_jxl`'s signature moves there, so this branch does not compile against the current `COUNTERPART_REV` pin.

`test_jxlsave` pinned a deferred contract, `encode_jxl` returning a typed `EncodeError::Unsupported`, and that no longer holds now JPEG XL has a decoder and a lossless encoder. The signature moves too: `encode_jxl(lossless: bool)` becomes `encode_jxl(jxl::SaveOptions)`, so this is a compile fix rather than only a behaviour follow-up.

The cell now does what the libvips cell it was ported from does, minus the two steps that have nothing behind them. Steps 2 and 4 there encode lossy and compare the buffer sizes; the only JPEG XL encoder reachable in pure Rust is `zune-jpegxl`, which is lossless modular with no VarDCT path anywhere in it, so `jxlsave`'s `distance`, `Q`, `tier` and `effort` have no spelling at all and there is no lossy buffer to size. That is the deliberate answer libviprs#620 took rather than accepting one of those arguments and discarding it, and the docstring says so where a reader will hit it.

Step 3 needs no tolerance, because there is no quantisation anywhere in the pipeline and the round trip is the exact identity. The cell checks it at both integer carriers, since JPEG XL holds 16-bit samples natively where WebP does not, and pins the float refusal as well, which is the one carrier vips writes and libviprs does not.

## Local run

Built against the core branch `feat/500-jxl` through the path dependency:

```
cargo fmt -- --check                                              clean
cargo test --features ported_tests --no-run                       builds
cargo test --features ported_tests --test ported_foreign -- test_jxlsave
  test test_jxlsave ... ok
  test result: ok. 1 passed; 0 failed; 0 ignored; 105 filtered out
```

`test_webp` was run alongside it and still passes, so the shared import block and the fixture path are both intact.

## Not changed

`COUNTERPART_REV` stays where it is. It gets the core merge SHA once libviprs#619 and libviprs#620 land, not before.

### DIFF
--- a/.commit-500-jxl-tests.txt
+++ b/.commit-500-jxl-tests.txt
@@ -1,0 +1,24 @@
+test(ported): follow JPEG XL encode to a real lossless codec
+
+libviprs#619 and #620 give JPEG XL a decoder and a lossless encoder, so the
+deferred contract test_jxlsave pinned, encode_jxl returning a typed
+EncodeError::Unsupported, no longer holds. The signature moves too:
+encode_jxl(lossless: bool) becomes encode_jxl(jxl::SaveOptions), so this is a
+compile fix rather than only a behaviour follow-up.
+
+The cell now does what the libvips cell it was ported from does, minus the two
+steps that have nothing behind them. Steps 2 and 4 there encode lossy and
+compare the buffer sizes; the only JPEG XL encoder reachable in pure Rust is
+lossless modular with no VarDCT path anywhere in it, so distance, Q, tier and
+effort have no spelling at all and there is no lossy buffer to size. That is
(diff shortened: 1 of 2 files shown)


spdrman:
Superseded by the branch below. A scratch commit-message file got staged into the first commit here, and this repo's ruleset refuses every push to a branch after the one that creates it, so the fix had to be a new branch rather than an amend.

PR #166, opened 2026-08-26
tools: let a repo's own local-ci.py drive its pre-commit hook (closes #165)

### LINKED ISSUE #165: install-hooks.sh writes a hardcoded command list and calls it an exact mirror of CI
`tools/install-hooks.sh` generates the pre-commit hooks for four repos from a hardcoded per-repo list of cargo commands:

```bash
LIBVIPRS_CARGO_STEPS=(
    "cargo clippy --all-targets -- -D warnings -W clippy::incompatible_msrv -W deprecated"
    "cargo clippy --all-targets --features pdfium -- -D warnings -W clippy::incompatible_msrv -W deprecated"
)
```

and the hook it writes announces itself as mirroring "this repo's CI Check & Lint job exactly".

For libviprs that is two of that job's seven steps, and nothing from the other four jobs: three of CI's twenty. Two places describe the same thing, nothing compares them, and `ci.yml` grew while the list did not. Tracked on the core side as libviprs/libviprs#657.

## What changes

libviprs/libviprs#658 adds `tools/local-ci.py` there, which reads `.github/workflows/ci.yml` at run time and runs what is actually in it, in a container carrying the toolchains the workflow asks for. It cannot drift, because it holds no copy of the commands.

So this generator should stop writing a list of its own for any repo that has one. The change is:

- If a repo ships `tools/local-ci.py`, write a hook that calls it (`--fast` for pre-commit, which is Check & Lint, MSRV and Docs; the slow Test and Integration jobs stay on push). The repo directory is now passed through so the generator can detect it.
- Otherwise keep the existing behaviour, because `libviprs-cli` and `pdfium-render` have no such script.
- Stop the fallback hook claiming it mirrors CI exactly. It says it is a subset now, and points at `local-ci.py` as the real thing, because a hook that overstates its coverage is worse than one that admits it.

Verified by running it: libviprs gets the `local-ci.py` hook, `libviprs-cli` gets the honest fallback, all four repos install without error.

spdrman:
`tools/install-hooks.sh` generates the pre-commit hooks for four repos from a hardcoded per-repo list of cargo commands:

```bash
LIBVIPRS_CARGO_STEPS=(
    "cargo clippy --all-targets -- -D warnings -W clippy::incompatible_msrv -W deprecated"
    "cargo clippy --all-targets --features pdfium -- -D warnings -W clippy::incompatible_msrv -W deprecated"
)
```

and the hook it writes announces itself as mirroring "this repo's CI Check & Lint job exactly".

For libviprs that is two of that job's seven steps, and nothing from the other four jobs: three of CI's twenty. Two places describe the same thing, nothing compares them, and `ci.yml` grew while the list did not. Tracked on the core side as libviprs/libviprs#657.

## What changes

libviprs/libviprs#658 adds `tools/local-ci.py` there, which reads `.github/workflows/ci.yml` at run time and runs what is actually in it, in a container carrying the toolchains the workflow asks for. It cannot drift, because it holds no copy of the commands.

So this generator should stop writing a list of its own for any repo that has one. The change is:

- If a repo ships `tools/local-ci.py`, write a hook that calls it (`--fast` for pre-commit, which is Check & Lint, MSRV and Docs; the slow Test and Integration jobs stay on push). The repo directory is now passed through so the generator can detect it.
- Otherwise keep the existing behaviour, because `libviprs-cli` and `pdfium-render` have no such script.
- Stop the fallback hook claiming it mirrors CI exactly. It says it is a subset now, and points at `local-ci.py` as the real thing, because a hook that overstates its coverage is worse than one that admits it.

Verified by running it: libviprs gets the `local-ci.py` hook, `libviprs-cli` gets the honest fallback, all four repos install without error.

Closes #165

### DIFF
--- a/tests/install_hooks_localci_selection.rs
+++ b/tests/install_hooks_localci_selection.rs
@@ -1,0 +1,177 @@
+//! Guards that `write_pre_commit` (`tools/install-hooks.sh`) actually picks
+//! between its two pre-commit templates on whether the target repo ships
+//! `tools/local-ci.py`: present, it gets a hook that defers to it; absent, it
+//! gets the hardcoded per-repo cargo list.
+//!
+//! `tests/install_hooks_mirror_ci.rs` exercises the hardcoded list's
+//! *contents* in detail, but its synthetic `Workspace` never lays down a
+//! `tools/local-ci.py` anywhere, so neither the detection logic nor the
+//! LOCALCI branch itself was exercised by anything before this — and
+//! `libviprs` ships `tools/local-ci.py` today, so that branch is live for
+//! every real contributor.
+//!
+//! Both branches are driven by actually running the installed hook, with a
+//! recording stand-in in place of whatever each branch shells out to, rather
+//! than by reading the hook's text (libviprs/libviprs#695).
+
+use std::process::Command;
+
+mod common;
+use common::hooks::{Workspace, make_executable};
+
+/// A stand-in for `tools/local-ci.py` that just proves it ran. Carries a
+/// real shebang so that invoking it directly by path (rather than through
+/// `python3 <path>`) also works, which matters for the two tests here that
+/// are not about the executable-bit fix and would otherwise fail for an
+/// unrelated reason under the pre-fix hook.
+const RECORDING_LOCALCI: &str = r#"#!/usr/bin/env python3
+import os
+with open(os.environ["LOCALCI_RECORD"], "a") as f:
+    f.write("LOCALCI-RAN\n")
+"#;
+
+/// A recording `cargo`, the same stand-in `tests/install_hooks_mirror_ci.rs`
+/// uses to let the hardcoded fallback branch run to completion without a
+/// real cargo project.
+const RECORDING_CARGO: &str = r#"#!/bin/sh
+printf 'cargo %s\n' "$*" >> "$CARGO_RECORD"
+exit 0
+"#;
+
+/// Run `repo`'s installed pre-commit hook with a recording `cargo` in front
+/// of it and a stand-in `tools/run_ported_cells.sh` beside it, and report
+/// what each side of the branch recorded: whatever `cargo` saw, and whatever
+/// `local-ci.py` saw.
+fn run_hook(ws: &Workspace, repo: &str) -> (String, String) {
+    let repo_dir = ws.repo(repo);
+    let hook = repo_dir.join(".git/hooks/pre-commit");
+    assert!(
+        hook.is_file(),
+        "no pre-commit hook was installed into {repo}, so this guard would \
+         pass on two empty recordings"
+    );
+
+    let bin = ws.root.join("cargo-recorder").join(repo);
+    std::fs::create_dir_all(&bin).expect("create the recorder directory");
+    let cargo = bin.join("cargo");
+    std::fs::write(&cargo, RECORDING_CARGO).expect("write the recording cargo");
+    make_executable(&cargo);
+
+    // The fallback hook reaches this one by relative path rather than
+    // through PATH (tests/install_hooks_mirror_ci.rs does the same).
+    let ported = repo_dir.join("tools/run_ported_cells.sh");
+    std::fs::create_dir_all(ported.parent().expect("tools dir")).expect("create tools dir");
+    std::fs::write(&ported, "#!/bin/sh\nexit 0\n").expect("write a stand-in run_ported_cells.sh");
+    make_executable(&ported);
+
+    let cargo_record = ws.root.join(format!("{repo}.cargo.record"));
+    let localci_record = ws.root.join(format!("{repo}.localci.record"));
+    let _ = std::fs::remove_file(&cargo_record);
+    let _ = std::fs::remove_file(&localci_record);
+    let path = format!(
+        "{}:{}",
+        bin.display(),
+        std::env::var("PATH").unwrap_or_default()
+    );
+
+    let out = Command::new("bash")
+        .arg(&hook)
+        .current_dir(&repo_dir)
+        .env("PATH", path)
+        .env("CARGO_RECORD", &cargo_record)
+        .env("LOCALCI_RECORD", &localci_record)
+        .output()
+        .expect("run the generated pre-commit hook");
+    assert!(
+        out.status.success(),
+        "the pre-commit hook for {repo} failed with every command stubbed to \
+         succeed:\n{}{}",
+        String::from_utf8_lossy(&out.stdout),
+        String::from_utf8_lossy(&out.stderr)
+    );
+
+    (
+        std::fs::read_to_string(&cargo_record).unwrap_or_default(),
+        std::fs::read_to_string(&localci_record).unwrap_or_default(),
+    )
+}
+
+/// A repo with no `tools/local-ci.py` gets the hardcoded per-repo cargo list,
+/// and nothing else.
+#[test]
+fn a_repo_without_local_ci_py_gets_the_hardcoded_fallback() {
+    let ws = Workspace::new();
+
+    let (cargo_ran, localci_ran) = run_hook(&ws, "libviprs-tests");
+
+    assert!(
+        !cargo_ran.is_empty(),
+        "libviprs-tests has no tools/local-ci.py, so its pre-commit hook must \
+         run the hardcoded cargo step list; nothing was recorded"
+    );
+    assert!(
+        localci_ran.is_empty(),
+        "libviprs-tests has no tools/local-ci.py, so its pre-commit hook must \
+         not have run one:\n{localci_ran}"
+    );
+}
+
+/// A repo that ships `tools/local-ci.py` gets a hook that defers to it
+/// instead of the hardcoded list.
+#[test]
+fn a_repo_with_local_ci_py_defers_to_it_instead() {
+    let ws = Workspace::new();
+
+    let local_ci = ws.repo("libviprs").join("tools/local-ci.py");
+    ws.commit(
+        "libviprs",
+        "add a recording stand-in for local-ci.py",
+        &[("tools/local-ci.py", RECORDING_LOCALCI)],
+    );
+    make_executable(&local_ci);
+    ws.install_hooks();
+
+    let (cargo_ran, localci_ran) = run_hook(&ws, "libviprs");
+
+    assert!(
+        !localci_ran.is_empty(),
+        "libviprs now ships tools/local-ci.py, so its pre-commit hook must \
+         defer to it; it never ran"
+    );
+    assert!(
+        cargo_ran.is_empty(),
+        "libviprs ships tools/local-ci.py, so its pre-commit hook must not \
+         also run the hardcoded cargo list:\n{cargo_ran}"
+    );
+}
+
+/// Low-severity fix, same review round: the hook used to invoke
+/// `local-ci.py` directly by path, relying on its shebang and the executable
+/// bit. A present-but-not-executable script failed with a confusing
+/// "Permission denied" while the hook's own error text claimed "CI will fail
+/// the same way," which was false for what was actually a local permissions
+/// problem. Invoking it through `python3` instead means the executable bit
+/// no longer matters at all — assert that directly with a stand-in that is
+/// deliberately left non-executable.
+#[test]
+fn the_localci_hook_runs_it_even_when_it_is_not_executable() {
+    let ws = Workspace::new();
+
+    ws.commit(
+        "libviprs",
+        "add a non-executable recording stand-in for local-ci.py",
+        &[("tools/local-ci.py", RECORDING_LOCALCI)],
+    );
+    // Deliberately not made executable: the fix invokes it via `python3
+    // "$REPO_DIR/tools/local-ci.py"`, so this must still run.
+    ws.install_hooks();
+
+    let (_, localci_ran) = run_hook(&ws, "libviprs");
+
+    assert!(
+        !localci_ran.is_empty(),
+        "a non-executable tools/local-ci.py did not run. The hook must \
+         invoke it in a way that does not depend on the executable bit or \
+         its shebang line (e.g. `python3 \"$REPO_DIR/tools/local-ci.py\"`)."
+    );
+}
--- a/tests/install_hooks_localci_worktree.rs
+++ b/tests/install_hooks_localci_worktree.rs
@@ -1,0 +1,184 @@
+//! Guards that the LOCALCI pre-commit hook (`write_pre_commit` in
+//! `tools/install-hooks.sh`) resolves `REPO_DIR` against the tree actually
+//! being committed in, not against the shared `.git/hooks` directory a
+//! repo's main checkout and every one of its linked worktrees have in
+//! common (libviprs/libviprs#684).
+//!
+//! This is the same bug `install_pre_push` already dodges for the pre-push
+//! hook, via `git rev-parse --show-toplevel`. The LOCALCI branch shipped
+//! with `REPO_DIR="$(cd "$(dirname "$0")/../.." && pwd)"` instead, which
+//! resolves from the hook's own path rather than from the tree it is
+//! running in, so it always pointed at the main checkout: a commit made
+//! inside a linked worktree silently ran `local-ci.py` against the wrong
+//! tree, with no error.
+//!
+//! Mirrors `tests/prepush_gate_tests_the_pushed_tree.rs`: install the hooks
+//! with the real installer, put a recording stand-in in place of
+//! `tools/local-ci.py` so nothing here asserts on the hook's text
+//! (libviprs/libviprs#695), create a linked worktree with
+//! `Workspace::worktree`, and run the installed hook the way git does — the
+//! hook file always lives under the main checkout's shared `.git/hooks`,
+//! while the process's working directory is whichever tree is actually
+//! being committed in.
+
+use std::path::Path;
+use std::process::Command;
+
+mod common;
+use common::hooks::{Workspace, make_executable};
+
+/// A stand-in for `tools/local-ci.py` that records its own resolved path
+/// (`sys.argv[0]`, made absolute) and its arguments, so a push can be judged
+/// by what the hook actually invoked rather than by reading the hook's text.
+/// Carries a real shebang so it also runs correctly when invoked directly by
+/// path under the pre-fix hook, isolating these tests to the REPO_DIR bug
+/// rather than entangling them with the separate executable-bit fix.
+const RECORDING_LOCALCI: &str = r#"#!/usr/bin/env python3
+import os
+import sys
+
+with open(os.environ["LOCALCI_RECORD"], "a") as f:
+    f.write("ARGV0=" + os.path.abspath(sys.argv[0]) + "\n")
+    f.write("ARGS=" + " ".join(sys.argv[1:]) + "\n")
+"#;
+
+/// Run the pre-commit hook installed into `repo`'s main checkout, with the
+/// process's working directory set to `cwd`. This is exactly the shape git
+/// itself uses: the hook file always lives in the main checkout's shared
+/// `.git/hooks`, and the working directory is whichever tree is actually
+/// being committed in — the main checkout itself, or one of its linked
+/// worktrees.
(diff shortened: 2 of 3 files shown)


spdrman:
**Nothing mandatory. Ships as-is.** One suggestion from the API panel: the generated hook now hard-requires a running Docker daemon at commit time for any repo shipping `tools/local-ci.py`, and on a machine with Docker stopped every commit fails until `--no-verify` while the message says "these are the real CI commands, so CI will fail the same way", which would be wrong. Worth a Docker preflight that falls back to the hardcoded subset rather than blocking the commit. Filed rather than held.

---

# Consolidated panel review: batch 2

Four experts reviewed these nine PRs independently, with no shared context and no knowledge of each other: Principal Software Architect, Principal Rust/Clang Systems Engineer, Rust & Clang Safety and Reliability Engineer, and API & Library Design Expert.

**All four returned shippable, with holds.** The holds are narrow and every one is measured rather than argued.

## What all four did, and what it settled

Each panel merged all eight PRs onto `main` in a throwaway clone and ran the composed tree. Unanimous result: **three conflict hunks, none semantic**, and the composed tree is green across every feature combination CI runs (`cargo test` 1777, `--features jxl` 1802, `--features svg` 1793, `--features object-store-sink` 1785, clippy clean under CI flags, MSRV clean). No repeat of batch 1's array-length problem.

The #660/#666 collision was deliberate, and three panels independently proved **both naive resolutions fail to compile**, in different ways:

| resolution | what breaks |
|---|---|
| take #666's side | `E0433: cannot find type AlphaOp`, because git's conflict region on #660's side contains the enum |
| take #660's side | `E0603: interpretation_max_alpha is private`, because `resample` imports it |

The splice, agreed by all three: **keep #660's `AlphaOp`, take #666's doc block and `pub(crate)`.** Both branches ship a byte-identical function body, which is what a genuinely shared concept looks like.

One panel flagged a quieter hazard the others missed: the `resample.rs` half of that conflict is **doc-comment only on both sides**, so resolving it backwards compiles, passes all 1,539 tests, and silently restores a false claim.

## The recurring defect, now at thirteen

The brief asked panels to weight one pattern: a check that cannot fail for the reason it claims to test. They found five more, four by **mutating a guard and watching 1,500+ tests stay green**:

- Deleting #659's post-read cap check: **invisible to the entire suite**. It is not dead code, and removing it turns a correct refusal into a silent truncation reported as "not a recognisable image".
- Reverting #666's affine half entirely: **1519 passed, 0 failed**. That half of the PR is unpinned.
- #666's one unsigned test uses a **constant raster**, so it passes under both the f64 and f32 spellings and cannot observe its own PR's change.
- #662's new guard, which exists to stop a fifth `n-pages` meaning arriving silently, cuts each file at the **first** `#[cfg(test)]`. In `engine.rs` that is line 8 of 4,374. A reviewer inserted a real `n-pages` writer immediately before the test module and the guard passed it.
- #661's fuzz-corpus check asserts `matches!(err, SourceError::Jxl(_))`, so any of nine variants passes, in the PR whose entire purpose is making those nine distinguishable. `src/exr.rs` already does this correctly with a per-file variant table and an exact count.

One panel's closing recommendation is worth adopting as policy: **if you cannot break a new guard and see a test go red, the guard is not finished.**

## Mandatory before merge

### #660, two independent holds

**The unsigned path regressed 42 to 72 percent.** Changing `alpha_map` from a generic `impl Fn` to a `fn` pointer defeated monomorphisation. Measured at 2048x1536, min of 4 runs at `opt-level=3`:

| | main | #660 | delta |
|---|---|---|---|
| `Rgba8` premultiply | 11.55 ms | 16.42 ms | +42% |
| `Rgba8` unpremultiply | 10.95 ms | 16.53 ms | +51% |
| `Rgba16` premultiply | 12.58 ms | 19.83 ms | +58% |
| `Rgba16` unpremultiply | 12.17 ms | 20.89 ms | +72% |

Confirmed in the disassembly: `nm` shows two `call_once` shims in the #660 binary that do not exist on main, and the loop body has a `blr x22` before the rounding. The irony is instructive: the **float** arm keeps its `match` inside the pixel loop and LLVM unswitches it perfectly, so the arm that looks unoptimised is the fast one, and hoisting the match into a `fn` pointer is what broke it. Two candidate fixes were measured; the const-generic form is preferred. The `Rgba16` remainder was **flagged as unexplained rather than asserted**, which is the right call.

**Five `try_` methods still panic on a float raster.** `try_recomb`, `try_stdif`, `try_bitand`, `try_bitor`, `try_bitxor` all reach `depth_max`'s panic with the exact message #660 was written to eliminate. Reachable on ordinary untrusted input:

```
decoded EXR format = FloatF32(4), interp = ScRgb
decode_file(exr).try_recomb  =>  PANIC
```

The crate already has `reject_float_input` and `ArithmeticError::FloatUnsupported`, used on six other paths. Add them, plus one property test asserting no `try_*` method panics on an `RgbaF32`, which closes the class rather than the instance.

Also: the CHANGELOG says "Nothing on that path changed" about the unsigned arm, and the interpretation stamp is new there. Measured, `Rgba16` tagged `Srgb` comes back tagged `Rgb16` on main and `Srgb` composed. The change is correct and matches `vips_premultiply` copying the header; the denial is not.

### #662, three holds

**Two `encode_tiff` doc blocks misdescribe the accessor, byte-identical to `main`.** `src/encode_tiff.rs:42-46` and `:718-721` promise the count travels back and that the last page is `get_n_pages() - 1`. After #662 that is false for any TIFF with 10,000 or more IFDs. The PR touches `encode_tiff.rs` and its diff contains **zero lines mentioning `get_n_pages`**. This is batch 1's defect repeating exactly: behaviour changed in an area carrying a contract note, and the note is not in the diff.

**`get_n_pages` deep-copies an unbounded blob to read an integer.** It goes through `get_field`, which returns an owned `MetadataValue` via `.cloned()`, and `MetadataValue` has a `Blob(Vec<u8>)` variant that the `.v` trailer restores from untrusted files. Measured with a 64 MiB blob under that key: **30.518 ms for 20 calls against 584 ns without**, a 52,000x cost on a `&self` accessor that returns `1`, in the `for page in 0..raster.get_n_pages()` idiom its own docs recommend. `MetadataFields::get` is `pub(crate)` in the same module; matching on it directly is one line and zero allocation.

**The new guard scans 8 lines of a 4,374-line file.** Cut at the **last** `#[cfg(test)]` immediately followed by `mod tests`, recurse into subdirectories, and assert every file yielded a cut of the expected shape so a new pattern fails loudly instead of shrinking the window. Better still, invert it: one `pub(crate) fn set_n_pages` and let the guard assert the key appears in exactly one file.

### #666, two holds

**The affine half is entirely unpinned.** Reverting `TapFetch::new`'s `alpha_max` to the old `SampleLayout` spelling passes 1519 tests. The path is live and interpretation-sensitive, and it has the **better** oracle of the two, since `affine.c:553` calls `vips_premultiply` directly. Measured on 8.18.6, the scRGB alpha-1.5 pixel at (1,1) through `premultiply | affine 0.5 | unpremultiply` is `(66.66667175292969, 13.333333969116211, 2.0, 1.0)`.

**Unsigned resize bytes move, and nothing tests it.** The `as f32` re-rounding applies to every carrier, not just float: 8 of 65536 `Rgba8` samples differ at `resize(2.0)`, 4 of 4096 `Rgba16` at `resize(0.5)`, max delta 1. Reverting only the two `as f32` spellings, leaving `bracket_max_alpha` in place, makes all four cases byte-identical to main, so 100% of the movement is the rounding. Whether the new bytes are closer to vips or further from it is unmeasured, which is the actual problem.

### #659, one hold

**The post-read cap check has no test and deleting it is invisible.** 1777 passed with the guard removed. It is reachable deterministically whenever `stat` reports a length smaller than what the file yields, a FIFO being the easy case, and removing it converts a correct refusal into a silent truncation handed to the decoder as a whole file. A `#[cfg(unix)]` FIFO test is about fifteen lines; assert the variant and `needed_bytes == cap + 1` so it cannot be confused with the stat-first refusal.

### #644, one hold, plus two already fixed

**The `#[allow(deprecated)]` is not nightly-only.** No `cfg`, no `cfg_attr`, no `build.rs`. It is unconditional on every toolchain and it is the only one in the crate, disarming `Cargo.toml`'s `deprecated = "deny"` on stable at exactly the call site that machinery was written to watch. Move it to `RUSTFLAGS="-A deprecated"` on the Miri step, or gate it behind a `build.rs` cfg.

Already addressed: **`timeout-minutes` is now set on all three merge-gate jobs.** The gate had failed at 1h27m, been cancelled at 1h41m, and run four hours without a verdict, and with no ceiling the only signal was "cancelled", which reads like somebody pressed stop rather than like the gate breaking. That is why four months of Miri being unable to compile went unnoticed. `-Zmiri-disable-isolation` remains unproven and the next run is what settles it.

### #658, both holds already fixed

**The platform guarantee held by accident.** Three panels disagreed here and the disagreement was the finding: one reported no `--platform` flag anywhere, another observed the container running x86_64. Both were right, because `DOCKER_DEFAULT_PLATFORM=linux/amd64` was set in the operator's shell. On any other machine the same command ran native arm64 and `Dockerfile.ci`'s `TARGETARCH` switch silently selected the `linux-arm64` pdfium build. Now pinned explicitly on both `docker build` and `docker run`, and validated: the container reports `x86_64`, and Check & Lint, MSRV and Docs all pass against the real derived command list.

**A skipped job reported itself as passed.** With no `libviprs-tests` sibling, the Integration job's script emitted `exit 0`, so the runner printed `PASS` for a job that compiled nothing and counted it toward "All jobs passed". A lane worktree has no sibling, so that was the default state. It exits 99 now, prints `SKIP`, and the summary refuses to claim everything passed when anything was skipped.

## Where the panels disagreed

**Whether to hold #660 and #666 at all.** The architect and the API reviewer both concluded "land them as a pair, in either order", reasoning that #666 without #660 leaves a documented workflow with no correct implementation. The systems and safety reviewers hold both, on measured defects the other two did not look for. **Resolution: fix the defects, then land as a pair.** The pairing argument is correct and independent of whether the code is right.

**How far to push the shared-ceiling abstraction.** The architect wants one `premultiply_max_alpha(raster)` carrying the whole rule including the unsigned branch, noting `composite.rs` documents the same oracle and measurably ignores the tag. The systems and API reviewers consider `interpretation_max_alpha` already the right abstraction and the differing call sites correct. **Resolution: take the shared helper as-is, file the composite divergence separately with its measurement.** Unifying the unsigned branches means picking between two functions that behave differently on a 4-byte carrier, which is its own decision.

**Whether the local CI mirror earns its keep.** All four say yes on the derive-from-`ci.yml` design and no on the claims as shipped. Both blocking claims are now fixed.

## Optional, filed rather than held

Grouped by where they landed, ranked roughly by value per unit of effort:

- **The `SniffedFormat` list is hand-maintained**, and that is exactly how `Jxl` escaped between #628 and #659: the test asserted "only these buffer whole" while `decodes_from_memory` already returned true for it, because the variant was missing from **both** lists and the arithmetic stayed consistent. A `const ALL` plus an exhaustive `match` makes the next one a compile error. Two panels raised this independently.
- **`try_smartcrop` still panics on float**, and #660 moves that panic from `arithmetic.rs:246` to `extract.rs:251` rather than removing it. One reviewer predicted #660 would fix it, measured, and corrected itself.
- **`read_file_bounded` uses infallible allocation** in the one function whose stated purpose is turning an abort into a returned error, in a crate with `try_reserve_exact` in nine other modules.
- **`Extend::White` inks from the depth, not the interpretation** (already filed as #667). One reviewer predicted this blows up an scRGB float raster and found it cancels when there is an alpha band, but not without one: 3-band float scRGB gives a max sample of 255 where black gives 0.75. Pre-existing.
- **Three ways to say "this format is not compiled in"**: `JxlError::FeatureNotEnabled`, and `Io(ErrorKind::Unsupported)` for svg and pdfium.
- **WebP is the last self-decoded codec without a typed error**, and its allocation ceiling still arrives as the collapsed `InsufficientMemory` that #661 exists to fix elsewhere.
- **Stringly-typed public error fields**: `what: &'static str` with three undocumented values documented as "e.g.", `expected: &'static str` with a closed set of two, `pixel_format: String` baking a foreign crate's `Debug` into a public field.
- **`decode_file` and `decode_bytes` have no `# Errors` section at all**, and `clippy::missing_errors_doc` is not enabled.
- **`max_pages` defaults to 100,000 against an accessor ceiling of 10,000**, so the crate generates values its own accessor is documented to reject.
- **Three CHANGELOG entries are filed under `Fixed` or `Changed` where this repo's own precedent is `Breaking`**: refusing previously-accepted input (#659), changing what a public accessor returns (#662), and moving output bytes on four public operations (#666).
- **The merge gate only fires on PRs into `release`**, so after #644 lands Miri still runs only when someone remembers to dispatch it. A weekly `schedule:` is one line.
- **`MemoryTracker::alloc` overflows** where `dealloc` saturates, adjacent to the only line #644 touches.
- **`FROM rust:latest` unpinned** in the CI image, so the container's `stable` freezes at first build while CI's tracks the real release.
- **`make ci` claims to mirror `ci.yml`** and runs 5 of its 20 steps, untouched by #658.

## Overall assessment

**Ship all nine, after the mandatory items.** They total perhaps a day across five PRs and none is speculative: every one was demonstrated by measurement, mutation, or disassembly.

The engineering underneath is strong, and the panels said so unprompted. #661's error enum is the most consistent addition to this crate's public surface in either batch, with field names and message grammar matching its four siblings exactly, and it genuinely closes the two-collapsed-ceilings defect with both halves pinned under mutation. #662's port of `vips_image_get_n_pages` was verified by compiling a C program against the installed libvips rather than reading the source. #660's float arithmetic reproduces vips bit-for-bit including the NaN and negative-zero edges, with tests written to fail if someone reverts the double-rounding. #666's interpretation table is built specifically to defeat the cancellation trap, with alpha values inside, above and far above each candidate ceiling.

What is weaker, consistently, is the checking rather than the code. Five of this batch's findings are guards that cannot fail, and four were found by the same cheap technique. That is the thing to carry forward.

Three reviewers corrected their own predictions mid-review and said so. That is worth more than the findings.

spdrman:
## Consensus review (4 independent passes: architecture, Rust systems, safety/reliability, API/design)

Rust-systems review had nothing in scope here (pure bash); the other three converged, two of them with an empirical repro, on the same critical bug.

### Mandatory

**1. The new hook resolves the wrong repo root inside a git worktree (Critical — independently reproduced by 2 of 3 relevant reviewers with an actual linked worktree, not just read from the diff).**

The new `LOCALCI` branch derives `REPO_DIR` from the hook script's own location:
```bash
REPO_DIR="$(cd "$(dirname "$0")/../.." && pwd)"
```
Git hooks live in the *shared* `.git/hooks` directory of the main checkout, common to every linked worktree. `$0` therefore always resolves to the main checkout, never to the worktree actually being committed in — verified directly: `pwd` and `git rev-parse --show-toplevel` correctly resolve to the worktree, `$0`-derived `REPO_DIR` does not. The hook then runs `local-ci.py --fast` against the *main checkout's* tree, silently, with no error.

This is the exact bug this same file already paid to fix once, for the sibling pre-push hook, under `libviprs/libviprs#684` — whose own committed fix comment says nearly word for word: *"A repo's main checkout and all of its linked worktrees share one hooks directory, and git invokes the hook with $0 inside that shared directory... which is how every lane worktree on epic #520 gated against `main` instead of against its own branch."* `install_pre_push` in this same file already dodges this correctly via `git rev-parse --show-toplevel`. The new pre-commit branch reintroduces the discarded approach right next to the comment warning against it, and directly contradicts this PR's own stated goal ("this hook and CI cannot disagree") for exactly the worktree-lane workflow this team uses.

Fix: `REPO_DIR="$(git rev-parse --show-toplevel)"`, matching `install_pre_push`'s existing pattern.

**2. The new branch has zero test coverage (Medium/High, corroborated by all 3 relevant reviewers).** `tests/install_hooks_mirror_ci.rs`'s synthetic `Workspace` never creates a `tools/local-ci.py`, so neither the detection logic nor (critically) the worktree-correctness of the new branch is exercised by anything. Given `libviprs` already ships `tools/local-ci.py` today, this branch is live in production for every real contributor and untested.

Fix: extend `Workspace`/add a fixture with a stand-in `local-ci.py`, and add a worktree-based regression test mirroring `tests/prepush_gate_tests_the_pushed_tree.rs` (reusing `Workspace::worktree()`) that would have caught finding #1 directly.

### Not blocking, will fix while in the area (Low)

- No executable check on `local-ci.py` before invoking it directly (unlike `install_pre_push`, which checks `[ -x ... ]` up front with a distinct diagnostic). A present-but-non-executable script currently fails with a misleading "CI will fail the same way" message for what's actually a local permissions problem.
- The hardcoded status string `"Running the CI job list locally (Check & Lint, MSRV, Docs)..."` is itself a hand-maintained guess at job names, the same drift-prone pattern this PR otherwise eliminates — cosmetic only, `--fast` itself derives its real behavior from `ci.yml` dynamically.

### Noted, deliberately not doing now (Medium, lower priority)

One reviewer flagged that depending on `local-ci.py --fast`'s existence and flag contract is an implicit, unversioned cross-repo coupling, unlike `COUNTERPART_REV`'s explicit SHA pin. Fair, but the blast radius is much smaller (a broken local hook fails loudly and is bypassable with `--no-verify`, unlike a wrong CI counterpart), and pinning a dev-convenience script's version would be disproportionate. Leaving as a documented-contract nice-to-have rather than blocking on it.

I'm implementing the mandatory fixes now and will push once the full gate is green again.

spdrman:
I fixed the worktree bug the consensus review flagged, plus the two Low items from the same round.

**Critical fix**: `tools/install-hooks.sh`, the LOCALCI heredoc in `write_pre_commit()`. Changed
```
REPO_DIR="$(cd "$(dirname "$0")/../.." && pwd)"
```
to
```
REPO_DIR="$(git rev-parse --show-toplevel)"
```
matching `install_pre_push`'s existing pattern right below it. Verified with an actual `git worktree add`: the old line always resolved to the main checkout regardless of which worktree the hook ran from, the new one resolves to whichever tree is live.

**New tests** (both install the hooks with the real installer and run the generated hook with a recording stand-in for `local-ci.py`, no text inspection):

- `tests/install_hooks_localci_worktree.rs`: creates a linked worktree, runs the pre-commit hook from inside it, and asserts `local-ci.py` ran from the worktree's tree, not the main checkout's. I temporarily reverted the `REPO_DIR` fix locally and confirmed both tests in this file go red for exactly that reason (main checkout's `local-ci.py` runs instead of the worktree's); reapplied the fix and they're green. Also covers two linked worktrees of the same repo staying distinct from each other.
- `tests/install_hooks_localci_selection.rs`: covers the branch selection itself, since the existing synthetic `Workspace` in `tests/install_hooks_mirror_ci.rs` never creates a `tools/local-ci.py` at all, so neither the detection logic nor the LOCALCI branch's existence was exercised by anything. A repo with `tools/local-ci.py` gets the LOCALCI hook, one without it gets the hardcoded fallback (both asserted by running the hook, not reading it).

**Low fixes**, same file:
- Executable check: invoking `local-ci.py` directly by path relied on its shebang and the exec bit, so present-but-not-executable failed with "Permission denied" while the message claimed "CI will fail the same way," which was false. Switched to `python3 "$REPO_DIR/tools/local-ci.py" --fast`, so the executable bit no longer matters at all. Covered by a dedicated test with a deliberately non-executable stand-in, confirmed red against the old direct-invocation code and green against the fix.
- Hardcoded job names: `"Running the CI job list locally (Check & Lint, MSRV, Docs)..."` was itself a hand-maintained guess. Changed to `"Running the fast CI jobs locally (tools/local-ci.py --fast)..."`.

**Verification**: `bash -n` and `shellcheck` clean on `tools/install-hooks.sh` and `tools/hooks/pre-push`; `cargo fmt -- --check` and `cargo clippy --all-targets -- -D warnings` clean; the four relevant test files (`install_hooks_mirror_ci`, `install_hooks_pre_push_slots`, plus the two new ones) all green, 12 tests total. Ran the full default `cargo test` suite too: one pre-existing failure in `tests/op_oversized_input_fallible_alloc.rs` (`project_matches_libvips_project_reference`, a `Uint32` vs `Gray16` container mismatch) that I confirmed is already present on this branch before my changes and unrelated to hooks, everything else passes.

Pushed as a normal fast-forward to `chore/hooks-run-real-ci`. Not merging, leaving that for the cross-PR composition check.

PR #169, opened 2026-08-27
fix(hooks): gate on the tree being pushed, not on the main checkout

### LINKED ISSUE #684: pre-push hook tests the main checkout, not the worktree being pushed, so lane gates cannot fail
`libviprs-tests/tools/run-tests.sh` resolves the tree under test from its own location, not from the tree being pushed:

```
88: WORKSPACE_ROOT="$(cd "$SCRIPT_DIR/../.." && pwd)"
92: LIBVIPRS_DIR="$WORKSPACE_ROOT/libviprs"
...
210: cd "$LIBVIPRS_DIR"
226: cd "$LIBVIPRS_DIR"
```

`$WORKSPACE_ROOT/libviprs` is the main checkout, always. So when `.git/hooks/pre-push` fires from a git worktree, the suite runs against whatever `main` happens to be, and never sees a single line of the branch being pushed.

Confirmed directly. Pushing from `.epic520/_wt/672-colourspace-alloc`:

```
worktree HEAD:      970211a      <- the code being pushed
main checkout HEAD: a356c50      <- the code actually tested
```

This whole epic runs its lanes in worktrees under `.epic520/_wt/`, so **every lane push where the hook was allowed to run has been gated against `main`**. The gate could not fail on the author's own changes. That is worse than having no hook, because a green pre-push reads as "my branch passes" and it means nothing of the kind.

It also explains something that had been read as good luck: the hook has never once caught a branch defect in this epic.

## What it should do

Take the tree from the push, not from the script's neighbours. The hook already knows it:

```
REPO_DIR="$(cd "$(dirname "$0")/../.." && pwd)"    # .git/hooks/pre-push, line 6
```

but it only uses that to locate `run-tests.sh` and then throws it away. Passing it through, and having `run-tests.sh` accept the libviprs path as an argument or environment variable with the current value as the default, is the smallest fix. A worktree's `.git` is a file rather than a directory, so `dirname` resolution needs a check against `git rev-parse --show-toplevel` instead of assuming a layout.

Worth doing at the same time as #683, since both are the same hook and the same script, but they are separate defects: #683 is that it is expensive and OOMs, this is that it tests the wrong code.

Found by the #672 lane. Part of #520.

spdrman:
`tools/run-tests.sh` worked out which trees to test from where the script itself
sits, so `$WORKSPACE_ROOT/libviprs` was the answer on every run. Epic
libviprs/libviprs#520 puts each lane in its own git worktree, and a repo shares
one hooks directory between its main checkout and every worktree hanging off it,
so `pre-push` fired from a lane and gated that branch against `main`. The gate
could not fail on the author's own changes, which is worse than having no gate at
all: green read as "my branch passes" and meant nothing of the kind.

**Finding the tree.** The hook cannot get there from `$0`, which points into the
hooks directory every worktree shares, and it cannot walk up from `.git`, which
in a worktree is a file holding a `gitdir:` pointer rather than a directory. It
asks `git rev-parse --show-toplevel`, the tree git runs the hook from, and
`--git-common-dir` for the main checkout beside it, which is what names the repo
and locates the sibling harness.

**Accepting it.** `run-tests.sh` had no parameter for the tree at all, so passing
one from the hook was not enough on its own. `--libviprs` / `--libviprs-tests`,
or `LIBVIPRS_DIR` / `LIBVIPRS_TESTS_DIR`, carry it now, with the old sibling
paths as the defaults so every existing caller gets exactly what it got before.

**Actually building it.** This is the part the issue does not reach, and on its
own it would have left the fix looking right and doing nothing. The Dockerfile
copies `libviprs/` and `libviprs-tests/` out of the context root, so the context
has to hold the two trees under those two names. It was the workspace root, which
only ever worked because those two names happen to sit there. The two trees get
staged into a scratch context instead. That also stops every other sibling in the
workspace going to the daemon on each build, which on this machine is around 150
lane worktrees of the same crate.

**Saying what it did.** A run prints the trees and their revisions before it
builds and again with the verdict, and `--plan` answers the same question in a
second without touching Docker. The second commit closes a leak the first real
push exposed: git exports `GIT_DIR` into every hook, and it beats `git -C`, so
the banner reported the pushing branch's HEAD as the revision of a tree it had
never looked at. That commit also stops `set -e` taking the script out on the
`docker run` line, which is why a SIGKILLed container came back as a bare
"Pre-push tests failed" with no verdict, no trees, and no cleanup.

## Evidence

Pushed this branch from a git worktree with the fixed hook installed, and looked
inside the container the gate built while it was still running:

```
$ docker exec libviprs-tests-run sha256sum /src/libviprs-tests/tools/run-tests.sh
f3bb48d20ead6910...  /src/libviprs-tests/tools/run-tests.sh

worktree  tools/run-tests.sh   f3bb48d20ead6910...   <- what it built
origin/main tools/run-tests.sh 2464ba58d63a2ca6...   <- what it used to build
```

`tests/prepush_gate_tests_the_pushed_tree.rs` exists only on this branch, and the
container has it too. Under the old script neither could have been in there.

The core half of it, pointed at the `#672` lane worktree that found this, and
read out of the running container:

```
$ docker exec libviprs-tests-run sha256sum src/raster.rs src/source.rs src/imageio.rs
b3cb9a03b532a148...  src/raster.rs
5ed58793d252998d...  src/source.rs
0746d0ea134272e2...  src/imageio.rs

                    worktree (b1f4120)   main checkout (8e189be)
src/raster.rs       b3cb9a03b532a148     993f77c95222cfcb
src/source.rs       5ed58793d252998d     3ba2b27d13edb3b0
src/imageio.rs      0746d0ea134272e2     72f3c0744c2389de
```

Three for three on the worktree, none on the main checkout. That is the exact
comparison the issue makes, the other way round.

That run then failed, on `arithmetic_large_input_fallible_alloc` with
`signal: 9, SIGKILL`, which is libviprs/libviprs#683 and the reason this branch
went up with `--no-verify`. Worth saying plainly: that is the first time the gate
has failed on a lane's own tree, because until now it could not.

Seven cells in `tests/prepush_gate_tests_the_pushed_tree.rs` pin both halves, so
a regression here fails out loud instead of going quiet again.

## Deploying it

The generated hook only reaches a repo when `tools/install-hooks.sh` runs there,
and this branch does not run it. Everyone wanting the fix needs one
`./libviprs-tests/tools/install-hooks.sh` from the workspace root after it lands.

libviprs/libviprs#683 is stacked on top of this branch.

Closes libviprs/libviprs#684

### DIFF
--- a/Dockerfile.dockerignore
+++ b/Dockerfile.dockerignore
@@ -1,17 +1,20 @@
 # BuildKit .dockerignore for the run-tests.sh Docker mirror.
 #
-# The mirror builds with the WORKSPACE ROOT (the parent of libviprs/ and
-# libviprs-tests/) as the build context and `-f libviprs-tests/Dockerfile`
-# (see tools/run-tests.sh). Docker therefore reads its ignore rules from the
-# context root's `.dockerignore`, NOT from `libviprs-tests/.dockerignore`.
-# BuildKit checks for a `<dockerfile>.dockerignore` next to the `-f` Dockerfile
-# first and prefers it, so this file (living in the libviprs-tests repo, in
-# scope) is the authoritative ignore set for the mirror build. It fully
-# replaces the context-root file, so it must carry every exclusion, not just
-# the native-library additions.
+# The mirror stages the two trees under test into a scratch context and builds
+# with that as the build context and `-f <tests tree>/Dockerfile` (see
+# tools/run-tests.sh, "Build context"). The `-f` Dockerfile therefore sits
+# outside the context it is building. BuildKit looks for a
+# `<dockerfile>.dockerignore` next to the `-f` Dockerfile first and prefers it
+# over a `.dockerignore` at the context root, and it does that whether or not
+# the two live in the same directory, so this file (living in the
+# libviprs-tests repo, beside the Dockerfile it belongs to) is the
+# authoritative ignore set for the mirror build. It fully replaces any
+# context-root file, so it must carry every exclusion, not just the
+# native-library additions.
 #
-# Patterns are relative to the build context (the workspace root); `**/` makes
-# them match under both libviprs/ and libviprs-tests/.
+# Patterns are relative to the build context, which holds exactly two entries,
+# `libviprs/` and `libviprs-tests/`, staged there by run-tests.sh. `**/` makes
+# them match under both.
 **/target/
 **/.git/
 **/.DS_Store
--- a/tests/common/hooks.rs
+++ b/tests/common/hooks.rs
@@ -1,0 +1,49 @@
+//! Reads the git hooks `tools/install-hooks.sh` writes, so the guards on them
+//! can assert against the text that actually lands in `.git/hooks/`.
+//!
+//! Two suites need this (`prepush_gate_tests_the_pushed_tree` for #684 and
+//! `prepush_gate_cost_controls` for #683) and both used to carry their own
+//! copy, including the heredoc marker string byte for byte. That marker is the
+//! fragile part: change the quoting in `install-hooks.sh` and every copy has to
+//! follow, and the one that does not follow fails with "no longer writes the
+//! pre-push hook from a heredoc" rather than with anything about the change
+//! that broke it. One copy, here.
+
+use std::path::{Path, PathBuf};
+
+/// This crate's root, which is also the repo root for the harness checkout.
+pub fn repo_root() -> PathBuf {
+    Path::new(env!("CARGO_MANIFEST_DIR")).to_path_buf()
+}
+
+/// Read a repo-relative file, naming the path if it is not there.
+pub fn read(rel: &str) -> String {
+    let path = repo_root().join(rel);
+    std::fs::read_to_string(&path).unwrap_or_else(|e| panic!("cannot read {}: {e}", path.display()))
+}
+
+/// The marker `install-hooks.sh` opens the pre-push heredoc with. Kept as one
+/// constant so the two guards cannot drift apart on it.
+const PRE_PUSH_HEREDOC: &str = "cat > \"$pre_push\" << 'HOOK'\n";
+
+/// The pre-push hook body exactly as `install-hooks.sh` writes it: the heredoc
+/// between `<< 'HOOK'` and the closing `HOOK`. Reading the generated text
+/// rather than the generator keeps the assertions pointed at what actually
+/// lands in `.git/hooks/pre-push`.
+pub fn generated_pre_push() -> String {
+    let script = read("tools/install-hooks.sh");
+    let start = script
+        .find(PRE_PUSH_HEREDOC)
+        .map(|i| i + PRE_PUSH_HEREDOC.len())
+        .expect(
+            "tools/install-hooks.sh no longer writes the pre-push hook from a \
+             `cat > \"$pre_push\" << 'HOOK'` heredoc; update PRE_PUSH_HEREDOC in \
+             tests/common/hooks.rs to follow it rather than deleting the guards \
+             that read it",
+        );
+    let rest = &script[start..];
+    let end = rest
+        .find("\nHOOK\n")
+        .expect("unterminated pre-push heredoc in tools/install-hooks.sh");
+    rest[..end].to_string()
+}
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -6,8 +6,9 @@
 //! Cargo treats `tests/common/mod.rs` (directory module) specially — it
 //! isn't compiled as its own integration target.
 //!
-//! The canonical-fixture loader sits in [`fixtures`], and [`workflows`]
-//! resolves the CI workflow files the pinning guards read. Below them: PDF /
+//! The canonical-fixture loader sits in [`fixtures`], [`workflows`]
+//! resolves the CI workflow files the pinning guards read, and [`hooks`]
+//! reads the hook text `tools/install-hooks.sh` generates. Below them: PDF /
 //! pdfium-specific shared helpers for the streaming-mode integration
 //! tests. The pdfium helpers are gated behind the `pdfium` feature so
 //! the module compiles cleanly with the default feature set.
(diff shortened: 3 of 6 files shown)


spdrman:
## Consensus review, batch 4

Four reviewers went over this independently. All four say merge. This is the cleanest PR in the batch and the fix is the right one.

### Verified positive

The `GIT_DIR` leak finding is real and it is the most valuable thing in the stack. git exports `GIT_DIR` into the hook and it beats `git -C`, which meant `fetch_reference_suite.sh` was running `git -C "$WORK" init` against the live repository and then failing on `remote add origin`. That failed every hook-driven run, so the ported cells were silently skipped on every pre-push, always. Unsetting the git environment is correct.

Asking git for the tree (`git rev-parse --show-toplevel`, `--git-common-dir`) rather than inferring from `$0` is the right resolution, and the evidence in the PR body is the good kind: sha256 of three files in the running container matching the worktree and not `main`, three for three.

One reviewer chased the load-bearing Docker claim underneath all of this, because it is the sort of thing that silently stops working: after the build context moved from the workspace root to a staged scratch dir, is `Dockerfile.dockerignore` still found, given the `-f` Dockerfile now sits outside the context? They built a probe with a `.dylib` in the context and the Dockerfile plus its dockerignore in an unrelated directory, and BuildKit did honour it. The `**/`-prefixed patterns also still match under the new two-directory shape. So the claim holds.

Also worth recording: the observation that the issue's own proposed fix would not have worked alone. Passing `LIBVIPRS_DIR` through changes nothing while the Dockerfile does `COPY libviprs/ libviprs/` out of a context rooted at the workspace. Staging into a scratch context is what actually fixes it, and it stops shipping every sibling worktree to the daemon as a bonus.

### Fold in before merge

`Dockerfile.dockerignore`'s header is now wrong. It says the mirror builds "with the WORKSPACE ROOT (the parent of libviprs/ and libviprs-tests/) as the build context" and that "patterns are relative to the build context (the workspace root)". This PR is what changed that. The patterns still work and a reviewer verified the lookup, but the explanation of why is stale, and this is the change that invalidated it.

`tests/prepush_gate_tests_the_pushed_tree.rs` defines `repo_root()`, `read()` and `generated_pre_push()`, and #170 adds a second file defining all three again, `generated_pre_push()` verbatim including the heredoc marker string. `tests/common/mod.rs` exists for this and documents the convention, and `counterpart_pinning.rs` uses it. The heredoc marker is the fragile part and there are now two copies to update when `install-hooks.sh` changes its quoting. A `tests/common/hooks.rs` settles it for both files.

`run_tests_still_defaults_to_the_sibling_layout` compares a logical `pwd` from `run-tests.sh` against `canonicalize()`, so a symlinked workspace layout makes the cell red. One reviewer hit it. `pwd -P` fixes it.

### Worth a follow-up, not this PR

The hook is a ~120-line heredoc inside a shell function, and three things follow from that: shellcheck sees a string literal, tests can only grep it rather than run it (both new test files do exactly that), and the fix only reaches a repo when someone re-runs `install-hooks.sh`, with no version stamp to make a stale install detectable.

The suggestion worth taking seriously is `core.hooksPath` pointed at a checked-in `tools/hooks/pre-push`. It fixes all three at once, and there is a bonus that goes to the root of #684: `core.hooksPath` resolves relative to each working tree's top level, so a linked worktree naturally runs its own copy and the shared-hooks-directory problem this PR works around stops existing rather than being worked around. I will file it.

Also for the follow-up pile: `stage_tree`'s rsync excludes are top-level only and `Dockerfile.dockerignore` does not name `.epic520`, so the epic scaffolding (41 MB here, four review worktrees) gets rsynced to the scratch context and shipped to the daemon on every run, with nested `target/` dirs inside it escaping rsync entirely and being caught only at build time after being copied. Small next to what this PR already removed, and one line. Worth a comment that `--exclude '/tmp/'` is what stops `stage_tree` recursing into its own destination, because that is load-bearing and currently unremarked.

### The cli gap

All four reviewers landed on `libviprs-cli` getting this hook with no slot in the suite, so a cli push gates against two trees that cannot contain the change. I filed it as #691 with the scoping worked out, and I want to be explicit that it is not a coverage hole: the differential skipping in the Docker gate is deliberate and documented in `CLI_CONTRACT.md` §7, and the dedicated `cli-differential` job guards it with `VIPRS_REQUIRE_CLI=1`, which turns a silent skip into a panic. What is left is that a cli developer pays for a full image build and gets an answer that cannot be about their push. Not this PR's job to fix, and #691 has three options costed.

spdrman:
## Review folded in

Three things, none of them a behaviour change. Pushed as `18ebd5b`.

**`Dockerfile.dockerignore`'s header.** It still described the build context as the workspace root, which this branch is what stopped it being. Rewritten to say what the context actually is now: a scratch directory holding exactly the two staged trees, with the `-f` Dockerfile living outside it, which is why the `<dockerfile>.dockerignore` lookup is the load-bearing part. The patterns are unchanged.

**`repo_root()`, `read()` and `generated_pre_push()` are in `tests/common/hooks.rs`.** Both this file and #170's `prepush_gate_cost_controls.rs` pull them from there now, so there is one copy of the heredoc marker instead of two, and it is a named constant (`PRE_PUSH_HEREDOC`) rather than a string literal repeated inside `find` and `len`. The panic message points at the constant, so a quoting change in `install-hooks.sh` gets told where to go.

**`pwd -P` in `run-tests.sh`.** I fixed this on the script side rather than the test side. The script was printing logical paths while staging physical ones, so the banner could name a symlink that is not what got built, and `run_tests_still_defaults_to_the_sibling_layout` was comparing that against `canonicalize()`. All five `cd … && pwd` sites resolve physically now and the test carries a note saying both sides are physical on purpose.

`cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` are clean, all 7 cells in `prepush_gate_tests_the_pushed_tree` pass, and the full default suite is green.

Not touched here, as agreed: the `core.hooksPath` rework (#695), the libviprs-cli slot (#691), and the `.epic520` rsync excludes.

Issue #56, opened 2026-07-06
[High] Remove the committed libpdfium.dylib and fetch a pinned, checksummed binary (libpdfium.dylib; Dockerfile:16-17; ci.yml:41)


spdrman:
## Problem
A 6 MiB provenance-less `libpdfium.dylib` is committed to git and is loaded via the cwd lookup during local runs. The Docker/CI harness fetches pdfium from floating/unpinned channels, so the certified library is not reproducible.

## Evidence
- `libviprs-tests/libpdfium.dylib` (committed); floating fetch `libviprs-tests/Dockerfile:16-17`; CI pin `ci.yml:41-47`.

## Proposed fix
Delete the committed dylib; fetch a pinned, checksum-verified pdfium binary (from `libviprs-dep`) in Docker and CI. (Pairs with the cwd-load fix in the core crate.)

## Acceptance criteria
- [ ] The committed dylib is removed from the repo.
- [ ] Docker/CI fetch a pinned, checksummed binary.

PR #170, opened 2026-08-27
fix(hooks): stop the pre-push gate OOMing, and stop it running for nothing

### LINKED ISSUE #683: pre-push hook runs the full Docker suite unconditionally and OOMs on a loaded machine
`.git/hooks/pre-push` runs the full Docker test suite through `libviprs-tests/tools/run-tests.sh` on every push. On this host it does not finish, and it fails in ways that look nothing like a test failure.

The #674 lane hit it twice:

- first attempt, `arithmetic_large_input_fallible_alloc` died with `signal: 9, SIGKILL`
- second attempt, the container was killed mid-compile with no message at all

Both are OOM kills rather than test failures, and the same jobs pass on GitHub runners. I hit the same hook from a different angle pushing #676: a two-file change touching only `.github/workflows/merge-gate.yml` and `CHANGELOG.md` kicked off a full `libviprs-tests:local` image build, pulled the rust and debian base layers, and was still going five minutes later.

Two separate problems, worth separating:

1. **It OOMs under load.** A machine already running several cargo builds does not have the headroom, and the hook reports SIGKILL or nothing at all, which reads as a broken test rather than as memory pressure. Either the Docker memory limit needs raising or `arithmetic_large_input_fallible_alloc` needs a cap on what it tries to allocate.
2. **It runs unconditionally.** A change that touches no Rust at all still triggers a full image build and suite run. A path filter, so a push that changes only workflows, docs or the CHANGELOG skips straight through, would remove most of the cost without weakening the check where it matters.

The current situation is worse than either fixing it or removing it, because `--no-verify` is documented in the hook itself as the escape hatch and the practical effect is that people push with it routinely... at which point the hook is protecting nothing while still costing everyone several minutes.

Lives in `libviprs-tests/tools/run-tests.sh` and `install-hooks.sh` rather than in this repo.

Part of #520.

spdrman:
Two separate reasons the gate cost more than it was worth. Between them, `git
push --no-verify` became the normal way to push, which leaves the hook
protecting nothing while still costing everyone several minutes.

## The OOM is a cell whose premise does not hold here

Not memory pressure, in the end. Both oversize-scratch cells assert that an
over-capacity scratch allocation is *refused*, and they get there by asking
`try_hough_circle` for a vote accumulator of roughly 137 TiB. That is refused
on macOS, which does not over-commit, and on the x86-64 runners, whose 47-bit
user address space (128 TiB) has nowhere to put the mapping, which is exactly
why the same cells are green on GitHub. On aarch64 Linux the 48-bit space
(256 TiB) has room: `try_reserve_exact` succeeds, `Vec::resize` starts writing
zeros into it, and the kernel kills the process. The local Docker mirror is the
one place that runs these on aarch64 Linux, so nothing else ever saw it.

Measured in the container, one test binary on its own:

```
PEAK-AFTER-COMPILE: 2154512384
running target/debug/deps/arithmetic_large_input_fallible_alloc-...
test hough_circle_oversize_scratch_panics_not_aborts ... EXIT=137
Killed
PEAK-AFTER-TEST:    3696230400     <- the ceiling
oom_kill 1
```

It is not a bug in libviprs. The size the cells pick is simply not
over-capacity on that host, so the fallible path they exist to exercise is
never reached.

So the cells stop asking the host what it will refuse and bring their own
ceiling: each re-runs itself in a child with `RLIMIT_AS` set well below the
request, where the reservation is refused everywhere, at once, with no page
touched. The contract and the sizes are unchanged. Same run, same container,
same `3525m`:

```
PEAK-AFTER-COMPILE: 2258055168
test result: ok. 3 passed; 0 failed
PEAK-AFTER-TEST:    2258055168     <- did not move
oom_kill 0
```

They also take a slot in turn, because they build the same 576 MB raster and
cargo runs the cells of one binary on a thread pool.

## The reporting around it was wrong in its own right

The ceiling was a hardcoded `--memory=4g`, and this host's Docker VM holds
3.8 GiB in total, so the cgroup never bound: the VM ran out first, the kernel
picked a victim inside the container, and `docker inspect` reported
`OOMKilled=false` because as far as the daemon was concerned nothing had hit
its limit. That is why these arrived as a bare `signal: 9, SIGKILL` under
whichever test name was unlucky, with nothing to say it was memory.

The ceiling is clamped under what the daemon actually has now, which puts a
kill back inside the container where it gets recorded, and a run that is killed
says so in those words with the three things worth trying. Concurrent rustc
processes come off the same budget, because rustc is what holds the memory
during the compile: around 130 test binaries, one per container CPU.
`RUN_TESTS_MEMORY`, `RUN_TESTS_BUILD_JOBS` and `RUN_TESTS_TEST_THREADS`
override any of it. The test threads keep cargo's own default, because the
Dockerfile explains why the pdfium suites want the full pool.

The integration step also compiles and runs in two calls with a `sync` between
them rather than in the one `cargo test` that does both. Building ~130 test
binaries leaves gigabytes of dirty page cache in the cgroup, and dirty pages
cannot be reclaimed until they are written back, so the first memory-hungry
binary to start could push the group past its ceiling faster than reclaim could
answer. That was what was left once the arithmetic cell stopped growing without
bound: the suite got as far as `streaming_pdfium_matrix` and died there instead.

With both in place the whole suite passes under the clamped ceiling, and **this
branch went up through the gate itself with no `--no-verify`**:

```
Green ported cells passed.
================================================================
All tests passed (arm64).
  libviprs:       /Users/rom/workspace/libviprs/libviprs (8e189be+dirty)
  libviprs-tests: /Users/rom/workspace/lv684/libviprs-tests (3024879)
```

(The one force-push after that carried a commit-message edit on a byte-identical
tree.)

## It ran unconditionally

The hook reads the refs git hands it, works out what the push actually changes,
and skips only where the answer is knowable.

## What it will not skip

Worth being explicit, because "touches no Rust" turned out not to mean "the
suite cannot see it":

| path | verdict | why |
| --- | --- | --- |
| `.github/**` in libviprs | skip | nothing in either repo reads it |
| `.github/**` in libviprs-tests | run | `tests/common/workflows.rs` reads `.github/workflows/*.yml`, and four pinning guards go through it |
| `README.md`, `CHANGELOG.md` | run | `tests/feature_rename_docs_present.rs` reads both, out of `../libviprs` as well as out of this repo |
| `docs/**`, `LICENSE`, `.gitignore`, `.gitattributes`, `.editorconfig` | skip | nothing reads them |
| anything else | run | |

So the two-file case in the issue, `merge-gate.yml` plus `CHANGELOG.md`, still
runs, on the CHANGELOG half. The workflow half is genuinely inert and the
CHANGELOG half genuinely is not, and I would rather the list stay defensible
than cover that one example. `LIBVIPRS_PREPUSH_ALL=1` runs the suite whatever
the paths say, and the hook names either the path that made it run or the paths
that let it skip, so the decision is never silent.

Driven through the generated hook in a synthetic workspace, with real ref
updates on stdin and a stub suite:

```
libviprs       .github/workflows/ci.yml               -> skipped
libviprs       .github/workflows/ci.yml + docs/x.md   -> skipped
libviprs       .github/workflows/ci.yml + README.md   -> ran (README.md)
libviprs       src/lib.rs                             -> ran
libviprs       workflow only, LIBVIPRS_PREPUSH_ALL=1  -> ran
libviprs-tests .github/workflows/ci.yml               -> ran (.github/workflows/ci.yml)
```

And the budget reports itself before anything is built:

```
budget: --memory=3525m, 2 cargo build job(s) (clamped: the Docker VM holds 3916m in total)
```

## Reading it

This branch sits on top of libviprs/libviprs#684 (PR #169), so its two commits
are in the diff below as well. Read #169 first, or read only the third commit,
`fix(hooks): stop the pre-push gate OOMing...`, which is all of this one.

It targets `main` rather than #169's branch because GitHub only registers a
closing keyword on a pull request that targets the default branch, and the two
issues are worth closing automatically. Merge #169 first and this one reduces to
its own commit.

Closes libviprs/libviprs#683

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,6 +62,15 @@ ENV CARGO_PROFILE_DEV_DEBUG=0
 # The wall-clock perf-ratio smoke is `#[ignore]`d here and runs in the nightly
 # workflow, so it never gates this container run.
 #
+# The integration step compiles and runs in two calls with a `sync` between
+# them, rather than in the one `cargo test` that does both. Building ~130 test
+# binaries leaves gigabytes of dirty page cache in the cgroup, and dirty pages
+# cannot be reclaimed until they have been written back, so the first
+# memory-hungry test binary to start could push the group over its ceiling
+# faster than reclaim could answer. The kernel then killed it, and it came back
+# as a SIGKILL against whichever test was running (libviprs/libviprs#683).
+# Flushing first costs a second and makes that cache cheap to give up.
+#
 # The final step runs the green ported_* cells (feature = "ported_tests") via
 # tools/run_ported_cells.sh, the single source of truth for the green-cell
 # list (deferred remainder: issue #77). The cells read the pinned libvips
@@ -82,7 +91,8 @@ CMD sh -c '\
     echo "================================================================" && \
     echo "Running libviprs-tests integration tests (with pdfium)..." && \
     echo "================================================================" && \
-    cd /src/libviprs-tests && cargo test --features pdfium && \
+    cd /src/libviprs-tests && cargo test --features pdfium --no-run && sync && \
+    cargo test --features pdfium && \
     echo "" && \
     echo "================================================================" && \
     echo "Running libviprs-tests green ported cells (ported_tests)..." && \
--- a/README.md
+++ b/README.md
@@ -152,10 +152,38 @@ repos (libviprs, libviprs-cli, libviprs-tests):
 - `cargo fmt -- --check` — rejects unformatted code
 - `cargo clippy --all-targets -- -D warnings` — rejects lint warnings
 
-**Pre-push** (runs on every `git push`):
-- Runs the full Docker test suite via `run-tests.sh`
+**Pre-push** (runs on `git push`, when the push can reach the suite):
+- Runs the Docker test suite via `run-tests.sh`, against the working tree being
+  pushed. A linked worktree gates on its own branch, not on the main checkout
+  (libviprs/libviprs#684).
+- Skips the run when nothing in the push can reach a test. The list of paths
+  that count as inert is deliberately short and partly per-repo: `.github/` and
+  `.gitignore` are inert for a libviprs push and *not* for a libviprs-tests
+  one, whose guards read both. `README.md` and `CHANGELOG.md` are on neither
+  list, because the documentation guards read them out of both checkouts. The
+  hook names the path that made it run, or the paths that let it skip.
+- `LIBVIPRS_PREPUSH_ALL=1 git push` runs the suite whatever the paths say. Use
+  this rather than `--no-verify` when you do not trust a skip decision: it is
+  the difference between overriding the filter and turning the gate off.
+
+Before pushing, `./tools/run-tests.sh --plan` answers "which trees, which
+revisions, what budget" in a second without touching Docker.
+
+The container budget is three environment variables, all read by
+`run-tests.sh` (see its `--help`):
+
+| Variable | What it does |
+|---|---|
+| `RUN_TESTS_MEMORY` | Hard memory ceiling, default `4g`, clamped to what the Docker VM actually has. A ceiling above the VM total never binds, and the kill then lands outside the container where the daemon does not record it. |
+| `RUN_TESTS_BUILD_JOBS` | Concurrent `rustc` processes, default derived from the ceiling at ~1.5 GiB each. This is the first lever on an out-of-memory kill. |
+| `RUN_TESTS_TEST_THREADS` | `cargo test` thread pool, default cargo's own. The Dockerfile explains why the pdfium suites want the full pool. |
+
+A container killed for memory says so in those words and prints the budget it
+used; it is not a test failure, whatever test name printed last.
 
 To bypass in emergencies: `git commit --no-verify` or `git push --no-verify`.
+For a pre-push you think is skipping wrongly, reach for `LIBVIPRS_PREPUSH_ALL=1`
+first, because `--no-verify` is how the gate stopped protecting anything.
 
 ## Test Suites
 
--- a/tests/arithmetic_large_input_fallible_alloc.rs
+++ b/tests/arithmetic_large_input_fallible_alloc.rs
@@ -34,19 +34,217 @@
 //! (vips-8.18.4) and committed under
 //! `tests/fixtures/arithmetic_large_input_fallible_alloc_expected/`.
 
-use libviprs::{PixelFormat, Raster};
+use libviprs::{ArithmeticError, PixelFormat, Raster, RasterError};
 
-/// A raster whose `hough_circle` vote accumulator is far larger than any host
-/// can address. A 24000×24000 Gray8 raster (~576 MB, well within the 8 GiB
-/// construction budget) with a full `1..=65535` radius range makes
-/// `acc = 24000 * 24000 * 65535 * 4 bytes ≈ 137 TiB`, above the ~64 TiB point
-/// at which the allocator refuses the reservation outright (rather than
-/// lazily over-committing it), so the failure is fast and deterministic: the
-/// pristine infallible `vec![0u32; …]` aborts, the fallible scratch returns
-/// `Err`.
+/// A raster whose `hough_circle` vote accumulator is far larger than this
+/// process is allowed to have. A 24000×24000 Gray8 raster (~576 MB, well
+/// within the 8 GiB construction budget) with a full `1..=65535` radius range
+/// makes `acc = 24000 * 24000 * 65535 * 4 bytes ≈ 137 TiB`.
+///
+/// "Larger than the process is allowed to have" is deliberately not the same
+/// claim as "larger than the allocator will hand out". 137 TiB is refused
+/// outright on macOS and on x86-64 Linux, and it is *not* refused on aarch64
+/// Linux, where the reservation fits in the 48-bit address space and the
+/// kernel kills the process partway through zeroing it. That difference is
+/// libviprs/libviprs#683 and it is why [`bounded_child`] exists: the cells run
+/// under an `RLIMIT_AS` of their own, so the refusal is a property of the test
+/// rather than a guess about the host. See its doc for the whole of it.
 const OVERSIZE_DIM: u32 = 24_000;
 const RADIUS_MIN: u32 = 1;
 const RADIUS_MAX: u32 = 65_535;
+
+/// A slot the two oversize cells take in turn.
+///
+/// Both build the same 24000x24000 Gray8 raster, ~576 MB apiece, and cargo
+/// runs the cells of one binary on a thread pool, so the binary's resident
+/// peak was twice what either cell needs. The slot is taken in the parent, so
+/// it also serialises the bounded children below.
+fn oversize_slot() -> std::sync::MutexGuard<'static, ()> {
+    static SLOT: std::sync::Mutex<()> = std::sync::Mutex::new(());
+    // A poisoned slot still serialises. The second cell panics on purpose, so
+    // refusing to hand the slot out after it would be the wrong answer.
+    SLOT.lock().unwrap_or_else(|poisoned| poisoned.into_inner())
+}
+
+/// Set in the child so it runs the cell body instead of spawning again.
+const BOUNDED_CHILD: &str = "LIBVIPRS_TESTS_BOUNDED_ADDRESS_SPACE";
+
+/// The child prints this before it execs the test binary, carrying whatever
+/// `ulimit -v` reports *after* the attempt to set it. The parent quotes it in
+/// every message, because "the child ran under a 2 GiB cap" and "the child ran
+/// uncapped and the allocator refused instead" are two different results and
+/// the run has to say which one it got. macOS returns `unlimited` here: its
+/// `ulimit -v` fails with `Invalid argument` and the cell falls back on the
+/// allocator refusal, which is exactly where it already was.
+const LIMIT_MARKER: &str = "address-space-kib:";
+
+/// The name of the enclosing `#[test] fn`, derived rather than typed.
+///
+/// The child is a re-exec of this binary with `--exact <name>`, and libtest
+/// exits 0 when a filter matches nothing (`running 0 tests` / `test result:
+/// ok. 0 passed`). A hand-typed literal that drifts from the function name
+/// therefore leaves both cells permanently green having executed nothing,
+/// which a reviewer demonstrated by changing one character. Nothing in the
+/// toolchain ties a string literal to a function name, so take the name from
+/// the function: `type_name` of a fn item declared inside it is
+/// `<krate>::<test fn>::probe`. [`bounded_child`] also insists on seeing
+/// `1 passed` come back, so if this ever resolves to something else the cell
+/// goes red rather than quiet.
+macro_rules! this_test_name {
+    () => {{
+        fn probe() {}
+        fn type_name_of<T>(_: T) -> &'static str {
+            std::any::type_name::<T>()
+        }
+        let path = type_name_of(probe);
+        let path = path.strip_suffix("::probe").unwrap_or(path);
+        path.rsplit("::").next().unwrap_or(path)
+    }};
+}
+
+/// 2 GiB, in KiB, which is what `ulimit -v` wants. Room for the 576 MB raster
+/// and the harness several times over, and nowhere near the ~137 TiB the vote
+/// accumulator asks for.
+const ADDRESS_SPACE_KIB: &str = "2097152";
+
+/// Re-run one oversize cell in a child process whose address space is capped,
+/// and return `false` in the parent once the child has been checked.
+///
+/// Both cells assert that an over-capacity scratch allocation is *refused*,
+/// and they get there by asking `try_hough_circle` for a vote accumulator of
+/// roughly 137 TiB. The premise underneath is that the allocator refuses a
+/// reservation that large outright. That holds on macOS, which does not
+/// over-commit, and on x86-64 Linux, whose 47-bit user address space (128
+/// TiB) has nowhere to put the mapping. It does not hold on aarch64 Linux:
+/// the 48-bit space (256 TiB) has room, `Vec::try_reserve_exact` succeeds,
+/// `Vec::resize` starts writing zeros into it, and the kernel kills the
+/// process somewhere past three gigabytes.
+///
+/// That is the `signal: 9, SIGKILL` in libviprs/libviprs#683, and it is not a
+/// bug in libviprs. The size these cells pick simply is not over-capacity on
+/// that host, so the fallible path they mean to exercise is never reached. It
+/// went unnoticed because the same cells pass on the x86-64 GitHub runners
+/// and on a macOS host, and the local Docker mirror is the one place that
+/// runs them on aarch64 Linux.
+///
+/// So the cells stop asking the host what it will refuse and bring their own
+/// ceiling. Each re-runs itself in a child with `RLIMIT_AS` set far below the
+/// request, where the reservation is refused everywhere, immediately, with no
+/// page ever touched. The contract under test is unchanged and the sizes stay
+/// where they are: the accumulator is still caller-scaled and still far past
+/// anything the process can have. What changes is that "past anything this
+/// process can have" is now a property of the test rather than a guess about
+/// the machine.
+///
+/// A host that ignores `ulimit -v` (Darwin does) lands back on the allocator,
+/// which is where these cells already were and where they already pass. If it
+/// ignores the limit *and* over-commits, the child dies rather than the whole
+/// test binary, and the parent says which cell and why instead of leaving a
+/// bare SIGKILL against a test name.
+fn bounded_child(test_name: &str) -> bool {
+    if std::env::var_os(BOUNDED_CHILD).is_some() {
+        return true;
+    }
+
+    let exe = std::env::current_exe().expect("path of the running test binary");
+    // `ulimit -v` is a no-op on Darwin and fails there with `Invalid argument`.
+    // Swallowing that and carrying on is right, the allocator refusal still
+    // holds, but the run has to say so, so read the limit back and print it
+    // before the exec. Output is captured rather than streamed, so the parent
+    // can both quote it and check that a test actually ran.
+    let out = std::process::Command::new("sh")
+        .arg("-c")
+        .arg(format!(
+            "ulimit -v {ADDRESS_SPACE_KIB} 2>/dev/null || :; \
+             echo \"{LIMIT_MARKER} $(ulimit -v)\"; \
+             exec \"$0\" \"$1\" --exact --test-threads=1"
+        ))
+        .arg(&exe)
+        .arg(test_name)
+        .env(BOUNDED_CHILD, "1")
+        .output()
+        .unwrap_or_else(|e| panic!("re-run {test_name} under a bounded address space: {e}"));
+
+    let stdout = String::from_utf8_lossy(&out.stdout).into_owned();
+    let stderr = String::from_utf8_lossy(&out.stderr).into_owned();
+    let regime = address_space_regime(&stdout);
+    let transcript = format!("--- child stdout ---\n{stdout}\n--- child stderr ---\n{stderr}\n---");
+
+    assert!(
+        stdout.contains(LIMIT_MARKER),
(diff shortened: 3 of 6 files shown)


spdrman:
## Consensus review, batch 4

Four reviewers went over this independently. This is the one PR in the batch I am holding, and not on style: there is a live defect on the skip list, and both of the guards meant to protect this change can pass while the thing they guard is gone. All three are provable in one edit and fixable in a dozen lines.

The diagnosis work underneath this PR is genuinely good and I want that on the record before the complaints. The OOM was never memory pressure, it was a cell whose premise does not hold on this architecture, and the 137 TiB arithmetic checks out (24000 x 24000 x 65535 x 4 B = 1.51e14 B = 137.3 TiB) with the 47-bit versus 48-bit split explaining the platform pattern exactly. Finding that `--memory=4g` was larger than the whole 3.83 GiB VM, so the cgroup never bound and the VM died first and `OOMKilled` came back false, is the kind of thing that stays mysterious for months. The failure reporting on the OOM path is the best in either repo: it says in plain words that a cgroup kill is not a test failure, prints the budget it actually used, and gives three ordered remedies cheapest-first. `--plan` is a real addition.

### Blocking 1: `.gitignore` is on the skip list, and the suite reads it

This is a live bug rather than a weak test. `inert_path()` has:

```sh
docs/*|.gitignore|.gitattributes|.editorconfig|LICENSE|LICENSE-*)
    true
    ;;
```

and `tests/pdfium_provenance.rs:181` reads it:

```rust
fn gitignore_excludes_native_binaries() {
    let ignore = read(".gitignore");
    for pat in ["*.dylib", "*.so", "*.dll"] {
        assert!(ignore.lines().any(|l| l.trim() == pat),
            ".gitignore must exclude {pat} so native binaries are never committed (issue #56)");
```

So a libviprs-tests push that changes only `.gitignore` skips the entire suite, and that is precisely the push that can break issue #56's guard against committing a provenance-less PDFium binary. The file's module doc says these tests "run on every `cargo test` invocation", which was true and now depends on a gate that decides whether `cargo test` runs at all.

A reviewer checked every other entry on that list against both trees and `.gitignore` is the only one anything reads: `docs/`, `LICENSE`, `.editorconfig` and `.gitattributes` are genuinely unread in both repos today, there is no `include_str!` of a `docs/` file anywhere, and the cross-repo reads (`../libviprs/src/colour.rs`, `../libviprs/Cargo.toml`, its README and CHANGELOG) are none of them inert. So the list is one entry from being right.

Condition `.gitignore` on `REPO_NAME` the way `.github/` already is, or drop it.

### Blocking 2: both cost-control guards can pass with the thing they guard removed

`skip_list_holds_nothing_the_suite_reads` is documented as "the skip list may only hold paths nothing reads". It checks five hard-coded literals. I lifted its logic and ran the real `inert_path` body through it with one poisoned entry at a time:

```
baseline (real list): PASSES

  add 'tools/*'      -> PASSES
  add 'src/*'        -> PASSES
  add 'Cargo.lock'   -> PASSES
  add 'tests/*'      -> PASSES
  add 'Dockerfile*'  -> PASSES
  add 'fixtures/*'   -> PASSES
```

`tools/*` is the one that matters, because `tools/run-tests.sh` **is** the suite, so exempting it means a change to the test runner never triggers a run. A reviewer went further and flipped `inert_path`'s fallthrough from `false` to `true`, which skips every push in every repo forever, and all four cells here plus all seven in `prepush_gate_tests_the_pushed_tree.rs` stayed green.

`the_skip_can_be_overridden` has the same shape. It asserts `hook.contains("LIBVIPRS_PREPUSH_ALL")`, the bare name, which appears three times in the hook but is only expanded once:

```
occurrences of the bare name: 3
occurrences of the expansion: 1

unmutated:            guard=True   hook_really_reads_var=True
escape hatch DELETED: guard=True   hook_really_reads_var=False
```

Delete the one line that reads the variable, keep the comment that mentions it, and the guard stays green with the escape hatch gone. Its own docstring describes that exact state as what it prevents.

Invert both into the set-equality shape #687 is using next door: run every tracked path in both repos through `inert_path` for real, collect what comes back inert, and assert it equals a pinned list. Then `tools/*` fails with a diff naming `tools/run-tests.sh`, and a widening needs two deliberate edits. For the override, assert the expansion (`${LIBVIPRS_PREPUSH_ALL`) at minimum, or drive the generated hook the way the PR body already did by hand.

Which brings up the thing worth fixing properly: the PR body has exactly the right evidence, six rows of hook-driven skip and run decisions, and it was run by hand and is not in the tree. For a change whose failure mode is the gate silently ceasing to run, that is the wrong thing to leave uncommitted. Roughly forty lines and no Docker needed.

### Blocking 3: `bounded_child` can pass having run zero tests

The child gets a hand-typed string duplicating the enclosing `#[test] fn` name, and the parent only checks `status.success()`. libtest exits 0 when a filter matches nothing, which I confirmed directly:

```
$ cargo test --test t -- --exact this_name_does_not_exist
running 0 tests
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 1 filtered out
EXIT=0
```

A reviewer changed one character in the name and inverted the body's assertion to `is_ok()`, and the cell still reported ok. So renaming either function silently disables both cells forever, and nothing in the toolchain ties a string literal to a function name. These are the only automated defence for a `handle_alloc_error` abort reachable from a caller-controlled radius range.

Use `.output()` and assert the child's stdout contains `1 passed` alongside the status check, printing the captured output on failure since `--nocapture` no longer streams. Deriving the name rather than typing it is better still.

### Fold in

`assert!(result.is_err())` at `:176` does not say which error, and `RLIMIT_AS` widens the set of things that return `Err` considerably. An early `if radii > K { return Err(...) }` added upstream would keep this green while it stopped exercising the accumulator. Assert the variant, which is the discipline #689 applies to all fourteen of its sites in this same batch.

Move `Raster::zeroed(...).expect(...)` out of `catch_unwind`. Under the 2 GiB cap a failed construction panics inside the closure, `caught.is_err()` is satisfied, and the cell passes without ever reaching `hough_circle`. Before this PR there was no cap and the construction could not plausibly fail, so this PR is what made it reachable.

The child cannot tell whether it was actually bounded. On macOS `ulimit -v 2097152` fails with `Invalid argument`, `2>/dev/null` eats it, and the `;` continues, so the child runs uncapped and proves the assertion by the old allocator refusal. I verified that on this host. The PR discloses it in prose, which is not the same as the run saying which regime it was in, and the whole point of this rework is that the old cells were green for a reason unrelated to what they claimed. Echo the limit before `exec`, or read it back with `getrlimit`.

The failure message diagnoses the opposite of the real regression. A reviewer reverted `try_hough_circle`'s accumulator to `vec![0u32; acc_len]`, the actual thing these cells exist to catch, and both died with SIGABRT, which is the allocator refusing and the infallible `vec!` aborting. The message says a kill means "the host honoured neither the cap nor the allocator refusal". Distinguish signal 6 from signal 9.

`OVERSIZE_DIM`'s doc still says the size is "above the ~64 TiB point at which the allocator refuses the reservation outright ... so the failure is fast and deterministic", which `bounded_child`'s own doc sixty lines later demonstrates is false on aarch64 Linux. Two contradictory explanations in one file, and the next person reads the first.

### Worth doing, not blocking

The `sync` is applied to one of three compile-then-run steps. If the dirty-page mechanism is real it applies to all three; if not, the change is a second fingerprint pass for nothing. Either split the others or say why the integration step is the only one that hits the ceiling.

`[ "$EXIT_CODE" -eq 137 ]` reports every SIGKILL as an OOM, including a `docker stop` or a Ctrl-C. In the hook, `CHANGED="$CHANGED\n$(git diff --name-only "$BASE" "$local_oid" --)"` is unguarded under `set -e`, so a `git diff` failure exits the hook and aborts the push with no message. `to_bytes` dies under `set -e` on a fractional `RUN_TESTS_MEMORY=1.5g` because `$(( 1.5 * 1024 ))` is an arithmetic syntax error; document integers only or parse with `awk`.

The gate's subject and its skip decision are different objects: `run-tests.sh` stages the working tree and `tree_desc` prints `+dirty`, while `inert_path` decides from the committed range being pushed. Commit a `docs/` change with a broken `src/lib.rs` uncommitted and the gate skips without looking at the source it would have staged. Forcing a run when `tree_desc` says `+dirty` closes it in one line.

`CARGO_BUILD_JOBS` is the right lever and the clamp is measured-true, but it does not bound a single rustc's peak, which on this crate plus moxcms is the likely tail: two jobs at 1.5 GiB against a 3.4 GiB ceiling leaves little for the link step. Worth naming `CARGO_PROFILE_DEV_CODEGEN_UNITS` in the escalation list so the next OOM is not diagnosed as "lower the job count" again.

`RUN_TESTS_CONTEXT_DIR` is the only one of the four `RUN_TESTS_*` variables missing from the header block `--help` prints, and `--dry-run` is an undocumented alias for `--plan`.

`CONTEXT_ROOT` sits beside the script so runs from different worktrees reuse a warm copy, which holds for a core push but not for a harness push, since the hook deliberately re-points `RUN_TESTS` at the pushed tree. Every harness lane then accumulates its own full copy of both trees. Either default it via `--git-common-dir` or correct the comment.

`libviprs-tests/README.md` still documents the old world: pre-push "runs the full Docker test suite" unconditionally, and `--no-verify` as the only named escape, which is the habit this PR exists to break. None of `--plan`, the three budget knobs, or `LIBVIPRS_PREPUSH_ALL` appears in it. That last one is surfaced in exactly two places, both inside a generated file, and one of them only prints on the skip path, so a developer learns the override exists only after the hook has already skipped for them.

The `LIBVIPRS_*` and `VIPRS_*` prefixes now split the same concept three ways: `LIBVIPRS_DIR`, `LIBVIPRS_TESTS_DIR` and `VIPRS_CLI_DIR` all answer "where does repo X live". This is the cheapest moment to settle it, since the new pair has two call sites and `VIPRS_*` has about a hundred.

### Verified positive

The cells do still catch the real regression: with the accumulator reverted to `vec![0u32; acc_len]` both go red, the child dies with SIGABRT, and the parent names the cell and the cap. Exit-status handling is right, `.status()` waits so there are no zombies, `sh -c` uses `"$0"`/`"$1"` so paths with spaces are safe, the `BOUNDED_CHILD` guard prevents recursive spawning, and `oversize_slot()` is taken in the parent so the children serialise and it is poison-tolerant. The 2 GiB cap has ample room over the 576 MB raster, so the "limit so low it is a different failure" hazard does not materialise. The `set -e` fix is complete for the failing-`docker run` path: `docker inspect` is `|| echo false`-guarded, `docker rm` is `|| true`-guarded, and all three verdict arms are reachable.

spdrman:
## Review folded in, with the mutations

All three blocking items and the fold-in list. The two rewritten guards now execute the real `inert_path` rather than grepping the text around it, and I ran your mutations against them. Evidence at the bottom.

### Blocking 1: `.gitignore`

You were right that it is a live defect and not a weak test. It moves onto the per-repo arm beside `.github/`:

```sh
.github/*|.gitignore)
    [ "$REPO_NAME" != "libviprs-tests" ]
    ;;
docs/*|.gitattributes|.editorconfig|LICENSE|LICENSE-*)
    true
    ;;
```

I re-checked the rest of the list the same way you did and got the same answer: `.gitignore` was the only entry anything reads, so the list is right now rather than one entry from right.

### Blocking 2: both cost-control guards

Inverted into the set-equality shape, and then some. `the_skip_list_exempts_exactly_the_pinned_paths` lifts the `inert_path()` function body verbatim out of the generated hook, wraps it in a driver, feeds it every tracked path in both repos (falling back to a working-tree walk in the container, where the context has no `.git`), and asserts the set that comes back inert equals a pinned list:

```rust
const INERT_IN_LIBVIPRS_TESTS: &[&str] = &["LICENSE"];
const INERT_IN_LIBVIPRS: &[&str] = &[
    ".github/workflows/ci.yml", ".github/workflows/merge-gate.yml",
    ".github/workflows/publish.yml", ".gitignore", "LICENSE",
    "docs/streaming-pdf-rotation.md",
];
```

Alongside it, `the_skip_list_never_exempts_what_the_suite_reads` asks the same function about twelve paths whose exemption would hurt most, so it names the consequence rather than the pattern and does not depend on what either tree happens to contain today.

And the behavioural evidence is in the tree now. `the_hook_skips_and_runs_the_way_the_list_says` builds a throwaway two-repo workspace, installs the whole generated hook in both, puts a stub in place of `run-tests.sh`, and drives seven real pushes past it. No Docker, and the whole file runs in under two seconds. The rows are yours, plus `tools/run-tests.sh`:

| repo | path changed | expected |
|---|---|---|
| libviprs-tests | `.gitignore` | run |
| libviprs-tests | `.github/workflows/ci.yml` | run |
| libviprs-tests | `LICENSE` | skip |
| libviprs-tests | `tools/run-tests.sh` | run |
| libviprs | `.github/workflows/ci.yml` | skip |
| libviprs | `src/lib.rs` | run |
| libviprs | `README.md` | run |

`the_skip_can_be_overridden` asserts the expansion `${LIBVIPRS_PREPUSH_ALL` and then drives the same workspace twice over the same LICENSE-only push, once plain and once with `LIBVIPRS_PREPUSH_ALL=1`, checking that the first skips and the second reaches the stub.

### Blocking 3: `bounded_child`

`.output()` instead of `.status()`, and it asserts the child's stdout carries `1 passed` alongside the status check, printing the whole captured transcript on any failure. The name is derived rather than typed:

```rust
macro_rules! this_test_name {
    () => {{
        fn probe() {}
        fn type_name_of<T>(_: T) -> &'static str { std::any::type_name::<T>() }
        let path = type_name_of(probe);
        let path = path.strip_suffix("::probe").unwrap_or(path);
        path.rsplit("::").next().unwrap_or(path)
    }};
}
```

Belt and braces on purpose: if `type_name` ever resolves to something else the filter matches nothing and the `1 passed` check goes red rather than quiet.

### The rest

**The child says which regime it ran in.** It echoes `address-space-kib: $(ulimit -v)` after the attempt and before the `exec`, the parent insists on seeing that line, and every message quotes what it said. On this macOS host the run now states in words that the cap did not apply and the assertion is resting on the allocator refusal, instead of that being true only in a doc comment.

**Signal 6 and signal 9 get separate sentences.** SIGABRT says `handle_alloc_error` fired on an infallible allocation and that is the regression the cells exist to catch; SIGKILL says nothing declined the request so it was served lazily and written into.

**`assert!(result.is_err())` names the variant**: `Err(ArithmeticError::Raster(RasterError::AllocationFailed { .. }))`, with a `describe()` for the message because `Raster`'s `Debug` prints the pixel buffer.

**`Raster::zeroed(..)` moved out of `catch_unwind`**, with a comment saying why this branch is what made it reachable.

**`OVERSIZE_DIM`'s doc** no longer claims the allocator refuses outright. It says 137 TiB is larger than the process is *allowed* to have, which is deliberately not the same claim, names the aarch64 case, and points at `bounded_child`.

**`run-tests.sh`:** a 137 that `docker inspect` does not record as an OOM gets its own arm now, naming both "memory, killed outside the cgroup's accounting" and "something stopped it: `docker stop`, Ctrl-C, the daemon restarting". `to_bytes` goes through awk, so `RUN_TESTS_MEMORY=1.5g` resolves to `1536m` instead of dying on `$(( 1.5 * 1024 ))` under `set -e`, and a value that is not a size is rejected by name. `RUN_TESTS_CONTEXT_DIR` is in the `--help` header, `--dry-run` is documented as the alias it already was, and `CARGO_PROFILE_DEV_CODEGEN_UNITS` is named in the escalation list so the next OOM is not diagnosed as "lower the job count" twice.

**The hook's `git diff`** is guarded, so a diff failure gives up on the skip decision (same answer as a base it could not find) rather than aborting the push under `set -e` with nothing printed.

**`README.md`** documents `LIBVIPRS_PREPUSH_ALL=1`, the three budget knobs and `--plan`, and says to reach for the override before `--no-verify`. That was your point about the override being surfaced in two places, both inside a generated file, one of which only prints once the hook has already skipped.

Not touched, as agreed: `core.hooksPath` (#695), the libviprs-cli slot (#691), the `.epic520` rsync excludes.

### Mutation evidence

Baseline on the branch: `prepush_gate_cost_controls` 7 passed, `arithmetic_large_input_fallible_alloc` 3 passed.

**Flip `inert_path`'s fallthrough `false` -> `true`** (the one that disables the gate for every push in every repo, and that all eleven cells used to survive):

```
the_skip_list_exempts_exactly_the_pinned_paths ... FAILED
the_skip_list_never_exempts_what_the_suite_reads ... FAILED
the_hook_skips_and_runs_the_way_the_list_says ... FAILED
test result: FAILED. 4 passed; 3 failed
```

**Add `tools/*` to the list**, same three fail, and the set-equality diff names the files:

```
the pre-push gate would now skip a push that changes only these 8 path(s)
in libviprs-tests, and they are not on the pinned list:
  tools/Dockerfile.libvips-pdfium
  tools/fetch_reference_suite.sh
  tools/gen_cli_expected.sh
  tools/gen_fixtures.sh
  tools/gen_pdfium_goldens.py
  tools/install-hooks.sh
  tools/run-tests.sh
  tools/run_ported_cells.sh
```

with the behavioural cell adding `a libviprs-tests push touching only tools/run-tests.sh must run the suite, because it is the suite`.

**Put `.gitignore` back on the unconditional arm** (i.e. undo blocking 1):

```
the pre-push gate would now skip a push that changes only these 1 path(s) in
libviprs-tests, and they are not on the pinned list:
  .gitignore

a libviprs-tests push touching only .gitignore must run the suite, because
tests/pdfium_provenance.rs reads this repo's .gitignore for the native-binary
patterns (issue #56), and a push that drops them is exactly the push that must
not skip. The hook said:
Pre-push: nothing in this push reaches the test suite, skipping it.
  .gitignore
test result: FAILED. 4 passed; 3 failed
```

**Delete the one line that reads the variable, keep the comment** (bare-name occurrences 3 -> 2, expansions 1 -> 0):

```
the pre-push hook mentions LIBVIPRS_PREPUSH_ALL but never expands it, so the
escape hatch is documentation only.
test result: FAILED. 6 passed; 1 failed
```

**Hand-type the child's test name one character wrong**, which is what the code did before:

```
the child exited 0 but no test ran: `--exact hough_circle_oversize_scratch_returns_typed_error_not_abor`
matched nothing, so this cell proved nothing.
--- child stdout ---
address-space-kib: unlimited
running 0 tests
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 3 filtered out
test result: FAILED. 2 passed; 1 failed
```

**Revert `try_hough_circle`'s accumulator to `vec![0u32; acc_len]`**, the regression the two cells exist to catch:

```
hough_circle_oversize_scratch_returns_typed_error_not_abort did not pass in a
child process (signal: 6 (SIGABRT)).
SIGABRT: the process aborted rather than returning. That is `handle_alloc_error`
firing on an infallible allocation, which is the regression these cells exist to
catch: the vote accumulator is not going through the fallible path. (the cap did
not apply (this host refused `ulimit -v`), so the child ran uncapped and the
assertion rests on the allocator refusing.)
--- child stderr ---
memory allocation of 150992640000000 bytes failed
test result: FAILED. 1 passed; 2 failed
```

Which is also the answer to your point about the message diagnosing the opposite thing: it now names SIGABRT and says what SIGABRT means here.

### The gate on its own tree

Pushed through the real hook, no `--no-verify`. All eleven cells pass inside the container on aarch64 Linux, which is the host the original SIGKILL came from:

```
Running tests/arithmetic_large_input_fallible_alloc.rs
test result: ok. 3 passed
Running tests/prepush_gate_cost_controls.rs
test result: ok. 7 passed
Running tests/prepush_gate_tests_the_pushed_tree.rs
test result: ok. 7 passed
...
All tests passed (arm64).
```

One data point for your "worth doing, not blocking" note on the budget. The first attempt, at the defaults, got all the way through both prepush binaries and the two oversize cells and then died in `streaming_pdfium_matrix`:

```
process didn't exit successfully: streaming_pdfium_matrix-… (signal: 9, SIGKILL: kill)

The container ran out of memory and the kernel killed it.
This is not a test failure, whatever test name printed last above.
The run had --memory=3525m and 2 cargo build job(s).
```

Which is the new reporting doing exactly its job, and also the shape of your `CARGO_BUILD_JOBS` point from the other direction: the residual peak on this host is at test runtime, not compile, so the escalation list's first lever is the wrong one for it. `RUN_TESTS_TEST_THREADS=2` cleared it and the run went green. I have left the escalation list alone rather than reorder it on one host's evidence, but it is worth someone measuring.

PR #184, opened 2026-08-29
test(project): expect the uint carrier the counting ops emit

### LINKED ISSUE #183: project_matches_libvips_project_reference asserts the 16-bit carrier project stopped emitting
`tests/op_oversized_input_fallible_alloc.rs` asserts `columns.format() == PixelFormat::Gray16` and reads the sums back as `u16`.

libviprs#532 moved the counting ops onto the `uint` carrier, because libvips emits every one of them as `VIPS_FORMAT_UINT` and a 300x300 image already overflows a 16-bit counter. `project` answers `Uint32(1)` now, so this test fails against libviprs `main`:

```
assertion `left == right` failed: project promotes column sums to a 16-bit container
  left: Uint32(1)
 right: Gray16
```

Measured on libviprs b06ba973, which is `main`, so this is not a lane's local state. It reproduces on this repo's own `main` (9450554).

## Fix

Expect `Uint32(1)` on both outputs, read the sums as `u32`, and widen the reference PNGs' 16-bit samples for the comparison. The fixture's sums fit 16 bits, so the reference stays exactly as captured rather than being re-taken: what moved is the container, not the arithmetic.

spdrman:
`project_matches_libvips_project_reference` asserts `Gray16` and reads `u16` sums. libviprs#532 moved `project` onto the `uint` carrier, so this repo's `main` is red against libviprs' `main`:

```
assertion `left == right` failed: project promotes column sums to a 16-bit container
  left: Uint32(1)
 right: Gray16
```

I found it running this suite against a libviprs branch and checked it reproduces at libviprs b06ba973 too, so it is not a lane's local state.

The reference PNGs are untouched. The fixture's sums fit 16 bits, so what moved is the container rather than the arithmetic, and widening the reference for the comparison says that instead of hiding it behind a re-capture. The row format is asserted now as well, which it was not before.

Verified passing against both libviprs b06ba973 and the branch I found it from.

Closes #183


spdrman:
## Consensus review (4 independent passes: architecture, Rust systems, safety/reliability, API/design)

I ran four adversarial reviews against this PR from separate perspectives with no cross-talk between them. Two findings converged independently across 3 of the 4 (the fourth, hooks-focused, had nothing in-scope here), which is why I'm treating them as mandatory rather than a single reviewer's opinion.

### Mandatory

**1. `tests/common/cli.rs`'s new signedness dispatch fights the crate's own purpose-built API for this (High, corroborated by all 3 Rust-literate reviewers).**

`max_abs_diff`/`max_abs_diff_interior` disambiguate `Int32` from `Uint32` with `matches!(a.format(), PixelFormat::Int32(_))` bolted onto a `(is_float, bytes_per_channel)` tuple match. `libviprs::PixelFormat` already ships `pub fn kind(self) -> SampleKind` specifically so callers don't do this — its own doc comment names this exact trap: *"Uint32, Int32 and FloatF32 are all four bytes and none of them is the others... That is why `PixelFormat::kind` rather than `PixelFormat::bytes_per_channel` is what to dispatch on when the question is how to read a sample (issues #516, #517, #607)."*

Concretely, this leaves `decode_compare`'s "format-class mismatch" assert unable to distinguish `Int32` from `Uint32` (both are `(false, 4)`), so a future carrier regression on a small non-negative fixture could pass silently instead of failing loudly, which is exactly the class of bug this whole PR exists to catch. It also duplicates ~15 lines of match logic across two functions, doubling the chance a future fix lands in only one.

Fix: replace the tuple-match-plus-guard in both functions with a single `match a.format().kind()`, extracted into one shared helper both call. Tighten `decode_compare`/`decode_compare_interior`'s format-class assert to compare `.kind()` (or full `PixelFormat` equality) instead of the `(is_float, bpc)` proxy.

**2. Stale doc comment (Medium, corroborated by 2/4 reviewers).**

`tests/op_oversized_input_fallible_alloc.rs`'s module doc still says the sums land in "a 16-bit container (saturating at `65535`...)" after this PR changes the assertion to `Uint32`. This file's whole subject is overflow/saturation safety, so a stale ceiling claim here is worse than average doc drift.

**3. `max_abs_diff`'s own rustdoc is now incomplete (Medium, api-design).** It still says "Reads samples as `u8`, `u16`, or `f32`," not mentioning the new `i32`/`u32` arms.

### Not blocking, will fix while in the area (Low, single-reviewer)

- `NonZeroU16::new(1).expect("one band")` could be `NonZeroU16::MIN` (a compile-time-obvious constant, no runtime-shaped construct needed).
- `use core::num::NonZeroU16` is inconsistent with `std::num::NonZeroU16` used elsewhere in the suite (e.g. `tests/ported_connection.rs`).

### Confirmed clean (no action)

All reviewers independently confirmed: no unsafe code or new OOB risk in the new byte-decode arms (bounds-checked slice indexing, existing width/height/channel asserts already guard the preconditions); the `csv_save()`/`matrix_save()` `.expect()` calls are honest, not just optimistic — both call sites are traced back to single-band (`channels() == 1`) inputs where the core's only fallible path (colourspace conversion) is structurally unreachable; the `COUNTERPART_REV`/`PROVENANCE.md`/`gen_cli_expected.sh` documentation updates are internally consistent and verified against the pinned upstream crate's actual doc comments.

I'm implementing the mandatory fixes now and will push once the full gate is green again.

PR #186, opened 2026-08-31
Bump COUNTERPART_REV: pick up with_kernel, with_vscale, Uint32

### LINKED ISSUE #185: COUNTERPART_REV pins a core from before with_kernel, with_vscale, and PixelFormat::Uint32 landed
`COUNTERPART_REV` still pins `f62a56a4`, and libviprs `main` has moved far enough past it that this suite's own CI has been red since 2026-08-28. Runs 33218089313, 33209570257, 33209320154, 33254125624, and 33312179143 are all `main`, all failure, and it isn't one job: Integration Tests, Integration Tests (pdfium), Feature Cell (packfile/jxl/object-store-sink/tracing), CLI Differential, and Lint all go down together, because they all compile the same test crate against the same stale counterpart.

The compile errors name two libviprs APIs that don't exist at the pinned commit, both already called from tests on this repo's `main`:

```
error[E0599]: no method named `with_kernel` found for struct `ResizeOptions` in the current scope
   --> tests/resample_nearest_alpha.rs:115:41
115 |     let opts = ResizeOptions::default().with_kernel(ReduceKernel::Nearest);
    |                                         ^^^^^^^^^^^ method not found in `ResizeOptions`
```

(run 33218089313, job "Integration Tests"). `tests/resample_premultiplied_alpha_reference.rs:186` calls `ResizeOptions::with_vscale` the same way. Neither method exists on `ResizeOptions` at `f62a56a4`.

It also blocks #184 (`fix/project-uint-carrier`, closes #183), which needs a third missing surface, `PixelFormat::Uint32`. That PR's own CI run (33277372556) fails for the identical reason:

```
error[E0599]: no variant, associated function, or constant named `Uint32` found for enum `PixelFormat` in the current scope
  --> tests/op_oversized_input_fallible_alloc.rs:95:23
95 |         PixelFormat::Uint32(one),
```

libviprs `main` is `694fd4b` as of this writing, 568 commits ahead of the pin. This is the same failure mode as #162 (COUNTERPART_REV pins a core from before the epic #520 batch landed), just a fresh recurrence: the pin falls behind, and every job that compiles against the newer surface goes red until someone bumps `COUNTERPART_REV` and re-verifies the suite against the new counterpart.

spdrman:
`COUNTERPART_REV` was pinned to `f62a56a4`, and libviprs `main` had moved 568 commits past it. Three of those commits are what force this bump: `ResizeOptions::with_kernel` and `ResizeOptions::with_vscale`, which `resample_nearest_alpha.rs` and `resample_premultiplied_alpha_reference.rs` already call on this repo's `main` (red since 2026-08-28), and `PixelFormat::Uint32`, which #184 needs (that PR is about issue 183).

I did not pin to libviprs `main` HEAD (`694fd4b`). Instead I bisected to `c07c685`, the mainline commit immediately before #899 (`fix/532-counters`) lands: it already carries all three required APIs, but not yet #899's change of `project`'s output carrier from `Gray16` to `Uint32`. Landing #899 breaks `op_oversized_input_fallible_alloc.rs`'s `project_matches_libvips_project_reference` on this repo's current `main`, which still asserts `Gray16`. #184 is exactly the fix for that carrier change, so I left #899 for whichever bump lands #184 rather than folding it into this one.

Getting even this far pulls in two feature landings that are unrelated to the three APIs above but structurally unavoidable, since both are ancestors of every commit that carries `with_kernel`: #754 makes UHDR save real (it now refuses non-float input with a typed `InvalidParameter` instead of always answering `Unsupported`), and a MATLAB `.mat` reader landed, so `decode_file` actually decodes `sample.mat` instead of erroring. Both broke `ported_foreign.rs` cells that pinned the old deferred-stub contract. I marked five of them `#[ignore]` with a precise reason each, rather than write new correctness assertions against features I don't have an independent oracle for: `test_matload`, `test_uhdrsave`, `test_uhdrsave_roundtrip`, `test_uhdrsave_roundtrip_hdr`, `test_uhdrsave_gainmap_scale_factor`. This is a call worth a second look, not a decision I'd call final: the reasons in each `#[ignore]` say exactly what real assertion should replace it.

Also fixed as a mechanical consequence of the newer counterpart: `tools/install-hooks.sh`'s libviprs step list was missing `jp2k`, `avif`, `packfile`, `serde`, and `tracing`, which the newer `ci.yml` already lints; `install_hooks_mirror_ci` caught it once the pin moved. And in `ported_foreign.rs`, `encode_jp2k` took a `(quality, lossless)` pair and now takes a `jp2k::SaveOptions`, and eight struct literals (four `gif::SaveOptions`, four `MagickLoadOptions`) stopped compiling once those types went `#[non_exhaustive]` with `with_*` builders; all nine call sites now use the builders.

## Verified locally

- `cargo test`: 774 passed, 0 failed, 11 ignored
- `cargo test --features "ported_tests tracing" --test phase3_tracing`: 12 passed
- `cargo test --features object-store-sink --test phase3_object_store_sink --test phase3_validation_stress --test sink_object_store_stub_contract`: 17 passed, 0 failed, 3 ignored
- `cargo test --features pdfium` (against a local pdfium build, `PDFIUM_PATH` pointed at a local dylib): 862 passed, 0 failed, 17 ignored
- `./tools/run_ported_cells.sh --clippy` and `--require-fixtures`: both green
- `cargo fmt -- --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo clippy --all-targets --features <feature> -- -D warnings` for `object-store-sink`, `packfile`, `tracing`, `jxl`: all clean
- `shellcheck tools/*.sh tools/hooks/pre-push`: clean

All of the above ran against the exact `c07c685` counterpart, laid out as a sibling checkout the way the integration job lays it out.

Closes #185

### DIFF
--- a/CLI_COUNTERPART_REV
+++ b/CLI_COUNTERPART_REV
@@ -37,4 +37,15 @@
 # (COUNTERPART_REV a055719), with the stale hough/rint/gamma doc comments and
 # OP_MAP.md rows refreshed. The bands/conversion/arithb/aritha differential
 # cells build and pass against this CLI + the fixed core.
-0988110bbd1d4fa9c4fc57279545bda088c246b4
+#
+# Bumped to libviprs-cli#47 @ 2235c11 (closes libviprs-cli#46), following
+# COUNTERPART_REV's move to c07c685: core's `libviprs#818` made every options
+# struct `#[non_exhaustive]` with `with_*` builders, and libviprs-cli's own
+# `src/ops/resample.rs:591` built `ResizeOptions` with a struct literal, which
+# stopped compiling the moment the core pin moved past `#818` — the
+# CLI-differential job cannot lay down a `viprs` binary at all without this,
+# regardless of which op family a given cell exercises. #47 confirmed it is the
+# only such construction site in the crate. Nothing about the command surface
+# or the reference fixtures changed, so every existing differential cell keeps
+# its previous classification.
+2235c1194f2e492fb5cbd9f30c22a32d566d4b31
--- a/COUNTERPART_REV
+++ b/COUNTERPART_REV
@@ -129,8 +129,35 @@
 # feature, so the ported cell here has two arms. #160 carries that cell and is
 # merged.
 #
+# Bumped to c07c685, not to libviprs main (694fd4b) as it stood when this bump
+# was made. Main had drifted 568 commits past the old pin, and three of those
+# additions forced the move: `ResizeOptions::with_kernel` and `with_vscale`
+# (resample_nearest_alpha.rs, resample_premultiplied_alpha_reference.rs, both
+# already on this repo's main and red since 2026-08-28) and `PixelFormat::Uint32`
+# (needed by #184, fixing #183). c07c685 is the mainline commit immediately
+# before #899 (fix/532-counters) lands: it already carries all three, but not
+# yet #899's change of `project`'s output carrier from `Gray16` to `Uint32`.
+# Landing #899 would break op_oversized_input_fallible_alloc.rs's
+# project_matches_libvips_project_reference on this repo's current main, which
+# still asserts `Gray16`; #184 is exactly the fix for that (closes #183), so
+# #899 rides along with whichever bump lands #184 rather than this one.
+#
+# Getting this far already dragged in two unrelated but unavoidable feature
+# landings, both ancestors of every commit that carries `with_kernel`: #754
+# (Ultra HDR save turned real, so it now refuses non-float input with a typed
+# `InvalidParameter` instead of always answering `Unsupported`) and the MATLAB
+# `.mat` reader (`decode_file` now actually decodes `sample.mat` instead of
+# erroring). Both broke ported_foreign.rs cells that pinned the old deferred
+# stub contract: test_matload, test_uhdrsave, test_uhdrsave_roundtrip,
+# test_uhdrsave_roundtrip_hdr, and test_uhdrsave_gainmap_scale_factor now carry
+# `#[ignore]` with a reason instead, since un-deferring them for real needs
+# fixture-specific verification this bump isn't the place for.
+#
 # Verified against this exact counterpart with the two checkouts laid out as
-# siblings, the way the integration job lays them out: the ported cells compile
-# under `--features ported_tests`, and `cargo test` gives 748 passed, 0 failed,
-# 11 ignored.
-f62a56a4e0487523e2e051d3aac6b7b84e2bb885
+# siblings, the way the integration job lays them out: `cargo test` gives 774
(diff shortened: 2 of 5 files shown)


spdrman:
Every job is green except CLI Differential, and I'm leaving that alone rather than routing around it.

It was already failing on main before this PR, for an unrelated reason (3 bicubic resample precision mismatches, e.g. `affine_bicubic_matches_vips_exactly`, visible on run 33218089313). This bump changes its failure mode from those 3 test failures to a full compile failure of `libviprs-cli`'s own `viprs` binary:

```
error[E0639]: cannot create non-exhaustive struct using struct expression
   --> src/ops/resample.rs:591:19
591 |       let options = ResizeOptions {
    |  ___________________^
592 | |         vscale,
593 | |         kernel,
594 | |         ..ResizeOptions::default()
595 | |     };
```

`ResizeOptions` went `#[non_exhaustive]` with `with_vscale`/`with_kernel` builders in the exact same libviprs commit this PR needs those two methods from, so there's no counterpart pin that gets this repo's own tests green without also breaking that struct literal. `libviprs-cli`'s `main` (pushed 2026-08-25) predates that change and has no commit that already uses the builders, so `CLI_COUNTERPART_REV` can't be bumped to a fix that doesn't exist yet. Fixing it means editing `libviprs/libviprs-cli` (a repo I don't have a mandate to touch here), following `CLI_CONTRACT.md` §7's bump order (core -> cli -> org -> tests), which this pin-bump is jumping ahead of on the cli leg.

Leaving this PR unmerged pending a decision on that.

spdrman:
Filed #187 to track replacing the 5 new `#[ignore]`s with real oracle-backed assertions (MATLAB .mat, UHDR). Backlog, not blocking this PR.

spdrman:
Update: bumped CLI_COUNTERPART_REV to `2235c11` (libviprs-cli#47, merged), which fixes the compile break this comment described. Re-ran the full CLI-differential coverage locally against both new pins: `cli_morphology_diff` plus all 17 op families the job exercises, 360 passed, 0 failed. That fix also cleared 5 pre-existing, unrelated resample precision failures (bicubic/vscale) as a side effect.

One new, unrelated failure turned up in the process: `hist_plot_golden_pin`. libviprs' own `2ba75e3` (an ancestor of every commit carrying `with_kernel`, so unavoidable at this pin too) changed `hist_plot` to emit `max` rows instead of `max+1`, specifically to match vips. The committed golden fixture is now stale (256x17 vs the real 256x16). Marked it `#[ignore]` with a reason rather than regenerate the fixture blind.

Checked the "org" step CLI_CONTRACT.md's bump order mentions: `libviprs-org` only publishes CLI reference docs, isn't wired into either repo's CI, and this fix didn't change any command surface, so nothing to do there.

Waiting on CI now.

Issue #202, opened 2026-09-11
[F2.3] PMTiles test suite: cross-backend equivalence, go-pmtiles interop, large-archive, CLI-driven


spdrman:
**Part of EPIC F (libviprs/libviprs#986) · Phase F2 · Issue 3/5** — _repo: `libviprs/libviprs-tests`_

## Objective
Build the black-box PMTiles test suite: cross-backend equivalence (directory vs PMTiles), independent `go-pmtiles` interop, large-archive/bounded-memory checks, and CLI-driven `viprs pyramid`/`viprs pmtiles …` tests — wired into CI with counterpart-pin bumps.

## Context
`libviprs-tests` is counterpart-pinned to exact SHAs (`COUNTERPART_REV`, `CLI_COUNTERPART_REV`; guarded by `tests/counterpart_pinning.rs`, `tests/cli_counterpart_pinning.rs`). It uses an offline-committed-reference oracle (`tests/common/dzsave_expected.rs`, `tests/common/cli.rs`, `tools/gen_fixtures.sh`) and decode-compares **pixels, never raw compressed bytes**. `tests/phase3_packfile.rs` is the archive-sink template. CI pins external binaries by curl+sha256 (pdfium, `.github/workflows/ci.yml:45-56`) — the exact pattern for `go-pmtiles`. The CLI harness builds `viprs` from the pinned counterpart (`tests/common/cli.rs`, `VIPRS_REQUIRE_CLI`).

## Scope
- `tests/phase_pmtiles.rs` (new): generate a pyramid into `fs://` and `pmtiles://`, decode both, and assert every logical `z/x/y` tile matches pixel-for-pixel via `assert_tiles_pixel_equal_tol` (`tests/common/dzsave_expected.rs`) — the critical cross-backend equivalence test.
- `go-pmtiles` interop:
  - **Offline golden** — commit a small reference `.pmtiles` (converted by `go-pmtiles`) under `tests/fixtures/pmtiles/` with a `PROVENANCE.md`; assert libviprs reads it tile-for-tile.
  - **Optional CI verify** — a job that downloads a pinned `go-pmtiles` release (curl + sha256, mirroring pdfium) and runs `go-pmtiles verify`/tile extraction against a libviprs-generated archive, gated by an availability/env check like `cli_available()` so dev machines skip cleanly.
- Large-archive/bounded-memory: a synthetic test exercising many tiles / large-offset reads asserting correct random access without whole-file reads.
[...]

PR #203, opened 2026-09-11
test(pmtiles): the black-box PMTiles suite, the CLI half, and both counterpart pins

spdrman:
Part of EPIC F (libviprs/libviprs#986), phase F2. Refs libviprs/libviprs-tests#202.

**This says `Refs` rather than `Closes`, and one acceptance criterion decides it.** #202 asks
for "the optional CI job verifies a libviprs-produced archive with `go-pmtiles`". The job is
here, five repo-file guards hold it in place, and the cell behind it passes against a real `pmtiles`
binary. But it passes against the **arm64** build, and the job downloads the **x86_64**
tarball. I have not downloaded that tarball, have not verified its digest and have not run that
job, so the only thing that can prove it is a hosted run, which happens after this merges.
Rounding that up would settle an acceptance criterion on an observation nobody has made. Everything
else on #202 is done and measured below, and one hosted run is all that is left of it.

Draft, and the epic owner merges it.

## What landed

| File | What it is | Cells |
|---|---|---|
| `tests/phase_pmtiles.rs` | Cross-backend equivalence: one plan into `FsSink` and `PmTilesSink`, both read through `PyramidReader`, compared as decoded pixels at tolerance 0. | 7 |
| `tests/pmtiles_interop.rs` | The three committed go-pmtiles goldens, the header cross-check, the leaf offset base, and the env-gated external-binary leg. | 8 |
| `tests/pmtiles_bounded.rs` | Offsets past 4 GiB, the counting `RangeReader`, the streaming memory bound. | 6 |
| `tests/cli_pmtiles.rs` | The CLI half: `viprs pyramid` with no output and no `--storage`, and the whole `viprs pmtiles` group over what it wrote. | 5 |
| `tests/pmtiles_ci_wiring.rs` | Repo-file guards on the interop job and on the CLI cell's wiring. | 6 |
| `tests/common/pmtiles.rs` | The oracle loader, the counting reader, the sparse archive, the materialiser. | |
| `tests/fixtures/pmtiles/` | Three goldens, two vector dumps, `PROVENANCE.md`. 30 KB. | |
| `.github/workflows/ci.yml` | The `pmtiles-interop` job, and `--test cli_pmtiles` on `cli-differential`. | |
| `COUNTERPART_REV`, `CLI_COUNTERPART_REV` | Both bumped to the merged EPIC F commits on the counterparts' own `main`. | |
| `tests/install_hooks_pdfium_scope.rs` | Nothing to do with PMTiles: the cells that were red on `main`. | +1 |

Both pins name a commit on `main`, not a branch head, and that distinction is the point rather
than pedantry. `.github/actions/clone-counterpart` fetches a raw SHA and checks out
`FETCH_HEAD`, so a branch head resolves perfectly well and then stops being an ancestor of
`main` the moment anything rewrites the branch, and it keeps resolving through
`refs/pull/N/head` after the branch is deleted. Nothing would ever fail: `tests/counterpart_pinning.rs`
only checks the SHA's shape and the clone action only checks that HEAD equals the pin. CI would
quietly test a tree nobody has. Confirmed at bump time rather than assumed, by asking
`gh api repos/<counterpart>/commits/main` for each one.

## The two red cells on main, which were not what they looked like

I was told two cells in `install_hooks_pdfium_scope` were failing, with a guess that a newer
clippy had grown a lint on a line the fixture leaves clean. That is a cause somebody stated
rather than one anybody measured, so I measured it, and it is not what is happening.

On the pinned CI image and on the newest `rust:latest` (both cargo 1.98.1, clippy 0.1.98),
`ptr_arg`'s primary span on that fixture is line 3, where the `&Vec<u8>` is. Not line 7.

What actually happens is that every fixture in that file builds a crate called
`pdfium-render-standin v0.0.0`, one per test, at its own temp path, and cargo does not keep
those apart when they share a `CARGO_TARGET_DIR`. I reproduced it on its own, outside this
suite, with two crates of that name and nothing else in them:

```
=== a first (shared target dir) ===
    Checking pdfium-render-standin v0.0.0 (/tmp/repro/a)
 --> src/lib.rs:3:21
 --> src/lib.rs:7:23
=== b second (same shared target dir) ===
 --> src/lib.rs:3:21
 --> src/lib.rs:7:23
=== b again with its own target dir ===
    Checking pdfium-render-standin v0.0.0 (/tmp/repro/b)
 --> src/lib.rs:3:21
```

`a` has `&Vec<u8>` on lines 3 and 7. `b` has `&Vec<u8>` on line 3 and `&[u8]` on line 7. Over a
shared directory `b` gets no `Checking` line at all: it is Fresh, never compiled, and it
replays `a`'s diagnostics, so a `&Vec` warning arrives pinned to a line of `b` that reads
`&[u8]`. Give `b` its own directory and the line-7 warning is simply not there. That is a lint
verdict about a file the compiler never looked at, and it is the `src/lib.rs:7` in the report.

`tools/local-ci.py` in the core repo exports `CARGO_TARGET_DIR` for every job, so the container
mirror hits this every time and a plain `cargo test` on a laptop does not. Local green, mirror
red, same code.

The fix is in the harness, not the hook and not the fixture: the stand-in gets a target
directory inside its own temp workspace, outside every stand-in repo so `git add -A` cannot
stage it. A real fork checkout has one crate of that name, and the hook honouring the user's
`CARGO_TARGET_DIR` is correct behaviour.

### The two cells, and what actually proves the fix

The decisive measurement is H1 in the table below. Take **this branch's** file, which is green,
and remove one line from it, the `CARGO_TARGET_DIR` override in `run_hook_full`, changing nothing
else. Four cells go red, and two of them are exactly the two I was told about:

```
a_change_that_introduces_no_lint_commits_over_inherited_debt          FAILED
an_ambient_deny_warnings_does_not_turn_inherited_debt_into_a_refusal  FAILED
a_deletion_that_breaks_the_build_is_refused                           FAILED
a_target_dir_shared_with_another_fork_does_not_decide_this_one        FAILED
```

Put the line back and all seven pass. That ties the reported failures to the missing target
directory and to nothing else, which no amount of reasoning about clippy versions could have
done.

The second measurement is `main`'s copy of the file, run verbatim in the same container with
`CARGO_TARGET_DIR` exported, against this branch's copy, four runs each:

| Copy | Run | Cells | Failed |
|---|---|---|---|
| `main`'s copy | 1 | 6 | a_change_that_introduces_a_lint_is_refused, a_deletion_that_breaks_the_build_is_refused |
| `main`'s copy | 2 | 6 | a_change_that_introduces_a_lint_is_refused, a_deletion_that_breaks_the_build_is_refused |
| `main`'s copy | 3 | 6 | a_change_that_introduces_a_lint_is_refused, a_deletion_that_breaks_the_build_is_refused |
| `main`'s copy | 4 | 6 | a_change_that_introduces_a_lint_is_refused, a_deletion_that_breaks_the_build_is_refused |
| this branch's copy | 1 | 7 | nothing |
| this branch's copy | 2 | 7 | nothing |
| this branch's copy | 3 | 7 | nothing |
| this branch's copy | 4 | 7 | nothing |

Two things in there are worth saying carefully, because it would be easy to overclaim.

`main`'s copy really is red here, so this is not a branch-only problem, and this branch's copy
really is green, four times out of four, with nothing flaky about it. But **the pair that reds
on `main` in my container is not the pair that was reported to me**. Mine is stable across all
four runs and it is a different two cells. I did not reproduce the reported pair directly and I
am not going to claim I did.

That difference is the interesting part rather than a loose end. A newer clippy growing a lint
would fail the same cells on every machine, because it would be a property of the compiler and
the fixture. What I have instead is three different failure sets from one cause: the pair
reported on the mirror, a different pair on my machine, and a four-cell set when the pinning is
removed from an otherwise-fixed file. Which cells lose depends on which same-named neighbour
reached the shared cache first, and that is decided by scheduling rather than by code.

## Mutation table

Every row takes its baseline from `git show HEAD:<path>` and refuses to run at all if the tree differs from HEAD. Every row asserts what the mutation **removed, by count, in the region the check parses**, before anything runs, rather than grepping that an edit landed. Every write is atomic with the original file mode kept and the mtime bumped, and is followed by a sha256 probe inside the container that has to report the bytes written before the expensive step starts. Every row restores from HEAD and then **runs a clean baseline**, and the sweep aborts if that baseline is not green, because a baseline that fails is the one signal separating "this mutation was caught" from "I am reading the last mutation's binary". Every run asserts that `Running` lines plus `Doc-tests` lines equals `test result:` lines, so a binary that fails to compile cannot quietly shorten a row.

All rows below were measured against the counterparts at the two SHAs this PR pins, not against a stand-in.

### The CLI half

Mutations in `libviprs-cli/src/main.rs`, which is what `tests/cli_pmtiles.rs` drives, and in the core function its index walk goes through.

| # | Mutation | File | What went red |
|---|---|---|---|
| C1 | derive the archive name with a different extension | `libviprs-cli src/main.rs` | `cli_pmtiles_info_reports_the_archive_the_library_sees`, `cli_pmtiles_tile_hands_back_the_bytes_the_tree_holds`, `cli_pmtiles_verify_accepts_what_pyramid_wrote_and_refuses_a_damaged_one`, `cli_pyramid_gives_the_same_pyramid_to_both_storages`, `cli_pyramid_writes_a_pmtiles_the_library_can_read` |
| C2 | hardcode JPEG tiles on the archive path only | `libviprs-cli src/main.rs` | `cli_pmtiles_info_reports_the_archive_the_library_sees`, `cli_pmtiles_tile_hands_back_the_bytes_the_tree_holds`, `cli_pyramid_gives_the_same_pyramid_to_both_storages` |
| C3 | print a constant addressed-tile count instead of the header's | `libviprs-cli src/main.rs` | `cli_pmtiles_info_reports_the_archive_the_library_sees` |
| C4 | drop the addressed-tiles cross-check, the one thing only the walk can see | `libviprs-cli src/main.rs` | `cli_pmtiles_verify_accepts_what_pyramid_wrote_and_refuses_a_damaged_one` |
| C5 | exit 0 after listing the problems | `libviprs-cli src/main.rs` | `cli_pmtiles_verify_accepts_what_pyramid_wrote_and_refuses_a_damaged_one` |
| C6 | swap x and y on the way into `get_tile` | `libviprs-cli src/main.rs` | `cli_pmtiles_tile_hands_back_the_bytes_the_tree_holds` |
| C7 | print the not-found diagnostic on stdout, keeping the non-zero exit | `libviprs-cli src/main.rs` | `cli_pmtiles_tile_hands_back_the_bytes_the_tree_holds` |
| C8 | transpose `tileid_to_zxy`'s answer, so the two directions of the mapping disagree | `libviprs src/pmtiles/tileid.rs` | `cli_pyramid_writes_a_pmtiles_the_library_can_read` |

### The library half

Mutations in `libviprs`, which `tests/phase_pmtiles.rs`, `tests/pmtiles_interop.rs` and `tests/pmtiles_bounded.rs` exercise through the public API.

| # | Mutation | File | What went red |
|---|---|---|---|
| M1p | drop the corner reflection from `rotate`, the wrong convention that hides in the structural rows | `libviprs src/pmtiles/tileid.rs` | `libviprs_reads_every_go_pmtiles_golden`, `the_duplicate_golden_serves_runs_as_well_as_direct_entries` |
| M2p | base a leaf's tile entry on `leaf_directories_offset` rather than `tile_data_offset` | `libviprs src/pmtiles/reader.rs` | `leaf_entries_resolve_against_tile_data_not_the_leaf_start`, `libviprs_reads_every_go_pmtiles_golden`, `the_four_gigabyte_boundary_itself_is_not_a_wrap`, `tile_offsets_past_four_gigabytes_resolve` |
| M3p | drop the run-end bound from `run_contains`, so landing on an entry counts as a hit | `libviprs src/pmtiles/directory.rs` | `a_coordinate_the_plan_does_not_cover_is_absent_from_both`, `a_tile_the_archive_does_not_hold_is_absent_rather_than_an_error`, `cli_pmtiles_tile_hands_back_the_bytes_the_tree_holds`, `tile_offsets_past_four_gigabytes_resolve` |
| M4p | truncate an entry's absolute position to 32 bits | `libviprs src/pmtiles/reader.rs` | `the_four_gigabyte_boundary_itself_is_not_a_wrap`, `tile_offsets_past_four_gigabytes_resolve` |
| M5p | mask an out-of-grid x and y the way go-pmtiles' own `ZxyToID` does | `libviprs src/pmtiles/tileid.rs` | `an_out_of_range_coordinate_is_refused_rather_than_masked` |
| M6p | swap `col` and `row` in `tile_coord_to_zxy` | `libviprs src/sink_pmtiles.rs` | `cli_pmtiles_tile_hands_back_the_bytes_the_tree_holds`, `cli_pyramid_writes_a_pmtiles_the_library_can_read`, `the_archive_addresses_tiles_at_the_coordinates_the_planner_named` |
| M7p | read the whole archive and slice it on every get_tile | `libviprs src/pmtiles/reader.rs` | `fetching_one_tile_reads_the_payload_and_not_the_archive`, `the_four_gigabyte_boundary_itself_is_not_a_wrap`, `tile_offsets_past_four_gigabytes_resolve`, `walking_every_tile_reads_the_payloads_and_nothing_else` |
| M8p | let the sink accept any layout, not only the ones addressed by (z, x, y) | `libviprs src/sink_pmtiles.rs` | `a_non_zxy_plan_is_refused_rather_than_relabelled` |
| M9p | silently drop every odd-column tile | `libviprs src/sink_pmtiles.rs` | `cli_pmtiles_tile_hands_back_the_bytes_the_tree_holds`, `cli_pyramid_gives_the_same_pyramid_to_both_storages`, `fetching_one_tile_reads_the_payload_and_not_the_archive`, `fs_and_pmtiles_agree_on_jpeg_tiles_too`, `fs_and_pmtiles_agree_tile_for_tile`, `go_pmtiles_accepts_and_reads_an_archive_libviprs_wrote`, `opening_an_archive_reads_the_header_and_the_root_and_stops`, `pmtiles_holds_every_tile_the_plan_declares`, `pmtiles_streaming_memory_bounded`, `the_archive_addresses_tiles_at_the_coordinates_the_planner_named`, `walking_every_tile_reads_the_payloads_and_nothing_else` |
| M10p | store every tile's bytes reversed | `libviprs src/sink_pmtiles.rs` | `cli_pmtiles_tile_hands_back_the_bytes_the_tree_holds`, `cli_pyramid_gives_the_same_pyramid_to_both_storages`, `cli_pyramid_writes_a_pmtiles_the_library_can_read`, `fetching_one_tile_reads_the_payload_and_not_the_archive`, `fs_and_pmtiles_agree_on_jpeg_tiles_too`, `fs_and_pmtiles_agree_tile_for_tile`, `the_archive_addresses_tiles_at_the_coordinates_the_planner_named` |
| M11p | report one zoom level more than the header has | `libviprs src/pyramid_reader.rs` | `both_backends_describe_the_same_pyramid` |

### The CI wiring, the fixtures and the hook mirror

Repo-file mutations. These need no counterpart at all, which is the point of keeping them in their own binary.

| # | Mutation | File | What went red |
|---|---|---|---|
| G1p | turn VIPRS_REQUIRE_GO_PMTILES off | `.github/workflows/ci.yml` | `the_interop_job_cannot_skip_to_a_false_green` |
| G2p | pin a different release from the one in the vector files | `.github/workflows/ci.yml` | `ci_pins_the_go_pmtiles_release_the_fixtures_were_produced_by` |
| G3p | drop the inner-binary digest check | `.github/workflows/ci.yml` | `ci_verifies_the_go_pmtiles_download_before_running_it` |
| G4p | drop `--test pmtiles_bounded` from the job | `.github/workflows/ci.yml` | `the_interop_job_runs_every_pmtiles_binary` |
| G5p | flip one byte of leaves-z0z7.pmtiles (offset 800, 0x47 to 0xb8) | `tests/fixtures/pmtiles/leaves-z0z7.pmtiles` | `a_tile_the_archive_does_not_hold_is_absent_rather_than_an_error`, `an_out_of_range_coordinate_is_refused_rather_than_masked`, `leaf_entries_resolve_against_tile_data_not_the_leaf_start`, `libviprs_reads_every_go_pmtiles_golden`, `our_header_decoding_matches_the_reference_field_for_field`, `the_committed_goldens_are_the_ones_the_vectors_describe` |
| G6p | rename dupes-z0z3.pmtiles out of the fixtures directory | `tests/fixtures/pmtiles/dupes-z0z3.pmtiles` | `a_tile_the_archive_does_not_hold_is_absent_rather_than_an_error`, `an_out_of_range_coordinate_is_refused_rather_than_masked`, `libviprs_reads_every_go_pmtiles_golden`, `our_header_decoding_matches_the_reference_field_for_field`, `the_committed_go_pmtiles_fixtures_are_all_present`, `the_committed_goldens_are_the_ones_the_vectors_describe`, `the_duplicate_golden_serves_runs_as_well_as_direct_entries` |
| G7p | drop the pmtiles-interop deferral from the hook's job list | `tools/install-hooks.sh` | `the_hook_mirrors_this_repos_ci` |
| G8 | drop `--test cli_pmtiles` from the cli-differential step | `.github/workflows/ci.yml` | `the_cli_pmtiles_cell_runs_where_it_cannot_skip` |
| G9 | rename the wired binary to `cli_pmtiles_extra`, so the substring is there and the boundary is not | `.github/workflows/ci.yml` | `the_cli_pmtiles_cell_runs_where_it_cannot_skip` |
| G10 | turn VIPRS_REQUIRE_CLI off on the step that runs it | `.github/workflows/ci.yml` | `the_cli_pmtiles_cell_runs_where_it_cannot_skip` |

### The stand-in hook harness

`tests/install_hooks_pdfium_scope.rs`, which has nothing to do with PMTiles and everything to do with the two cells that were red on `main`.

| # | Mutation | File | What went red |
|---|---|---|---|
| H1 | drop the CARGO_TARGET_DIR override from run_hook_full | `tests/install_hooks_pdfium_scope.rs` | `a_change_that_introduces_no_lint_commits_over_inherited_debt`, `a_deletion_that_breaks_the_build_is_refused`, `a_target_dir_shared_with_another_fork_does_not_decide_this_one`, `an_ambient_deny_warnings_does_not_turn_inherited_debt_into_a_refusal` |
| H3 | restore the whole file to its state on this repo's main (25ca6bc) | `tests/install_hooks_pdfium_scope.rs` | `a_change_that_introduces_a_lint_is_refused`, `a_deletion_that_breaks_the_build_is_refused` |

### The pins

P2 is the positive control for P1. A zero has two explanations and this is the one place in the sweep that produced one.

| # | Mutation | File | What went red |
|---|---|---|---|
| P2 | drop one character off the pinned SHA (the positive control for P1) | `COUNTERPART_REV` | `counterpart_rev_pins_a_full_sha` |
| P1 | point the pin at a 40-hex SHA that names no commit anywhere | `COUNTERPART_REV` | **NO TEST REDDENS**, see below |

<details><summary>Per-row mechanics: what each mutation removed by count, how many looks the container needed before it saw the bytes, and the baseline after the restore</summary>

```
C1    counts: 'input.with_extension("pmtiles")': 1 -> 0, replacement present 1x
      mount settled after 1 look(s); mutated run exit 101, 1 Running + 0 Doc-tests vs 1 result lines; baseline after restore failed=nothing exit 0
C2    counts: 'PmTilesSink::try_new(path, plan.clone(),': 1 -> 0, replacement present 1x
      mount settled after 2 look(s); mutated run exit 101, 1 Running + 0 Doc-tests vs 1 result lines; baseline after restore failed=nothing exit 0
C3    counts: 'println!("Addressed tiles: {}", header.a': 1 -> 0, replacement present 1x
      mount settled after 2 look(s); mutated run exit 101, 1 Running + 0 Doc-tests vs 1 result lines; baseline after restore failed=nothing exit 0
C4    counts: '    if walk.addressed_tiles != header.ad': 1 -> 0 (removed)
      mount settled after 2 look(s); mutated run exit 101, 1 Running + 0 Doc-tests vs 1 result lines; baseline after restore failed=nothing exit 0
C5    counts: '        for problem in &walk.problems {\n': 1 -> 0, replacement present 1x
      mount settled after 2 look(s); mutated run exit 101, 1 Running + 0 Doc-tests vs 1 result lines; baseline after restore failed=nothing exit 0
C6    counts: 'reader.get_tile(args.z, args.x, args.y)': 1 -> 0, replacement present 1x
      mount settled after 2 look(s); mutated run exit 101, 1 Running + 0 Doc-tests vs 1 result lines; baseline after restore failed=nothing exit 0
C7    counts: '        Ok(None) => operational_error(&f': 1 -> 0, replacement present 1x
      mount settled after 2 look(s); mutated run exit 101, 1 Running + 0 Doc-tests vs 1 result lines; baseline after restore failed=nothing exit 0
C8    counts: 'Ok((z, tx as u32, ty as u32))': 1 -> 0, replacement present 1x
      mount settled after 2 look(s); mutated run exit 101, 10 Running + 0 Doc-tests vs 10 result lines; baseline after restore failed=nothing exit 0
M1p   counts: '        if rx == 1 {\n            let mas': 1 -> 0 (removed)
      mount settled after 1 look(s); mutated run exit 101, 10 Running + 0 Doc-tests vs 10 result lines; baseline after restore failed=nothing exit 0
M2p   counts: '                let at = self.entry_posi': 1 -> 0, replacement present 1x
      mount settled after 1 look(s); mutated run exit 101, 10 Running + 0 Doc-tests vs 10 result lines; baseline after restore failed=nothing exit 0
M3p   counts: '            Some(end) => tile_id < end,': 1 -> 0, replacement present 1x
      mount settled after 1 look(s); mutated run exit 101, 10 Running + 0 Doc-tests vs 10 result lines; baseline after restore failed=nothing exit 0
M4p   counts: '        section_offset\n            .chec': 1 -> 0, replacement present 1x
      mount settled after 1 look(s); mutated run exit 101, 10 Running + 0 Doc-tests vs 10 result lines; baseline after restore failed=nothing exit 0
M5p   counts: '    if u64::from(x) >= side || u64::from': 1 -> 0, replacement present 1x
      mount settled after 1 look(s); mutated run exit 101, 10 Running + 0 Doc-tests vs 10 result lines; baseline after restore failed=nothing exit 0
M6p   counts: '    Ok((z, coord.col, coord.row))': 1 -> 0, replacement present 1x
      mount settled after 1 look(s); mutated run exit 101, 10 Running + 0 Doc-tests vs 10 result lines; baseline after restore failed=nothing exit 0
M7p   counts: '        Ok(Some(self.source.read_range(a': 1 -> 0, replacement present 1x
      mount settled after 1 look(s); mutated run exit 101, 10 Running + 0 Doc-tests vs 10 result lines; baseline after restore failed=nothing exit 0
M8p   counts: 'matches!(layout, Layout::Xyz | Layout::G': 1 -> 0, replacement present 1x
      mount settled after 1 look(s); mutated run exit 101, 10 Running + 0 Doc-tests vs 10 result lines; baseline after restore failed=nothing exit 0
M9p   counts: '        let (z, x, y) = tile_coord_to_zx': 1 -> 1, replacement present 1x; /map_err\(SinkError::PmTiles\)\?;\n\n        // Blank tiles/ (what the check parses): 1 -> 0
      mount settled after 1 look(s); mutated run exit 101, 10 Running + 0 Doc-tests vs 10 result lines; baseline after restore failed=nothing exit 0
M10p  counts: '        let bytes = self.encode(tile)?;\n': 1 -> 0, replacement present 1x
      mount settled after 1 look(s); mutated run exit 101, 10 Running + 0 Doc-tests vs 10 result lines; baseline after restore failed=nothing exit 0
M11p  counts: '            max_level: u32::from(header.': 1 -> 0, replacement present 1x
      mount settled after 1 look(s); mutated run exit 101, 10 Running + 0 Doc-tests vs 10 result lines; baseline after restore failed=nothing exit 0
G1p   counts: '          VIPRS_REQUIRE_GO_PMTILES: 1': 1 -> 0, replacement present 1x
      mount settled after 2 look(s); mutated run exit 101, 10 Running + 0 Doc-tests vs 10 result lines; baseline after restore failed=nothing exit 0
G2p   counts: '          GO_PMTILES_VERSION: 1.31.2': 1 -> 0, replacement present 1x
      mount settled after 2 look(s); mutated run exit 101, 10 Running + 0 Doc-tests vs 10 result lines; baseline after restore failed=nothing exit 0
G3p   counts: '          echo "${GO_PMTILES_BINARY_SHA2': 1 -> 0 (removed)
      mount settled after 2 look(s); mutated run exit 101, 10 Running + 0 Doc-tests vs 10 result lines; baseline after restore failed=nothing exit 0
G4p   counts: ' --test pmtiles_bounded': 1 -> 0 (removed)
      mount settled after 2 look(s); mutated run exit 101, 10 Running + 0 Doc-tests vs 10 result lines; baseline after restore failed=nothing exit 0
G5p   counts: byte 800: 47 -> b8, 1 byte(s) differ in 869
      mount settled after 2 look(s); mutated run exit 101, 10 Running + 0 Doc-tests vs 10 result lines; baseline after restore failed=nothing exit 0
G6p   counts: tests/fixtures/pmtiles/dupes-z0z3.pmtiles -> tests/fixtures/pmtiles/dupes-z0z3.sweep-hidden (file absent from the directory)
      mount settled after 1 look(s); mutated run exit 101, 10 Running + 0 Doc-tests vs 10 result lines; baseline after restore failed=nothing exit 0
G7p   counts: "                pmtiles-interop 'the tes": 1 -> 0 (removed)
      mount settled after 1 look(s); mutated run exit 101, 10 Running + 0 Doc-tests vs 10 result lines; baseline after restore failed=nothing exit 0
G8    counts: ' --test cli_pmtiles': 1 -> 0 (removed)
      mount settled after 1 look(s); mutated run exit 101, 10 Running + 0 Doc-tests vs 10 result lines; baseline after restore failed=nothing exit 0
G9    counts: '--test cli_pmtiles': 1 -> 1, replacement present 1x; /--test cli_pmtiles(?=\s|$)/ (what the check parses): 1 -> 0
      mount settled after 1 look(s); mutated run exit 101, 10 Running + 0 Doc-tests vs 10 result lines; baseline after restore failed=nothing exit 0
G10   counts: '      - env:\n          VIPRS_REQUIRE_CLI': 1 -> 0, replacement present 1x
      mount settled after 2 look(s); mutated run exit 101, 10 Running + 0 Doc-tests vs 10 result lines; baseline after restore failed=nothing exit 0
H1    counts: '    cmd.env("CARGO_TARGET_DIR", standin_': 1 -> 0 (removed)
      mount settled after 1 look(s); mutated run exit 101, 1 Running + 0 Doc-tests vs 1 result lines; baseline after restore failed=nothing exit 0
H3    counts: whole file replaced with 25ca6bc:tests/install_hooks_pdfium_scope.rs; 'CARGO_TARGET_DIR': 10 -> 0; 'standin_target_dir': 4 -> 0
      mount settled after 1 look(s); mutated run exit 101, 1 Running + 0 Doc-tests vs 1 result lines; baseline after restore failed=nothing exit 0
P2    counts: 'c1efa5651067b0cd02109a27a8eb8122496d0c40': 1 -> 0, replacement present 1x
      mount settled after 1 look(s); mutated run exit 101, 10 Running + 0 Doc-tests vs 10 result lines; baseline after restore failed=nothing exit 0
P1    counts: 'c1efa5651067b0cd02109a27a8eb8122496d0c40': 1 -> 0, replacement present 1x
      mount settled after 1 look(s); mutated run exit 0, 10 Running + 0 Doc-tests vs 10 result lines; baseline after restore failed=nothing exit 0
```

</details>

### The rows that started green, and the cells I had to write for them

Three mutations reddened nothing on a first pass. Two of them now have a cell of their own, and
the third has an explanation rather than a cell, which I would rather say plainly than paper
over.

**M6p, the transposed archive.** Swapping `col` and `row` in `tile_coord_to_zxy` reddened
nothing at first. Not the cross-backend comparison, not the count control, not the interop leg.
The reason is the good property turning round and biting: `PmTilesSink` and
`PmTilesPyramidReader` both map `TileCoord` through that one function, which is what stops them
drifting apart, and it is exactly why a mistake inside it cancels on a comparison that goes
through it on both sides. Nothing goes out of range either, because the grid check is against
`2^z` rather than the level's own cols and rows, so every swapped coordinate is addressable.
go-pmtiles and our raw reader then agree with each other about the transposed archive, because
neither has an opinion about what the planner meant. The archive is quietly transposed and a
consumer fetching `z/x/y` over HTTP gets the wrong tile.
`the_archive_addresses_tiles_at_the_coordinates_the_planner_named` asks `pmtiles::Reader::get_tile`
for the raw `(z, x, y)`, the one path that does not go through the shared function, and compares
against the file the directory backend put at that same `z/x/y`. It uses a deliberately
non-square plan, because on a square level a transposition is still a bijection onto itself and
half the coordinates land on a tile that happens to be there.

**C4, the `verify` walk nothing was testing.** The first version of
`cli_pmtiles_verify_accepts_what_pyramid_wrote_and_refuses_a_damaged_one` damaged the archive by
truncating it, and a truncation is caught when the archive is **opened**, before the walk starts.
Measured: with `verify`'s problem list neutered so it can never report anything, the truncated
archive was still refused and the cell stayed green, so nothing in the file was testing the walk
at all. The second damage is `addressed_tiles_count` set one too high: the archive parses, every
directory decodes, every tile reads, and the only thing wrong with it is a number only `verify`
compares. C4 removes exactly that comparison and the cell now reddens.

**H2, the cell that passed under the bug it was written for.** The first version of
`a_target_dir_shared_with_another_fork_does_not_decide_this_one` passed with the fix removed.
Cargo decides freshness by comparing source mtimes against the artifact, and that version built
the clean fork *after* the dirty one had already compiled, so the clean one was newer, was
rebuilt, and reported its own lints correctly. There was nothing left to replay by the time it
mattered. The suite's own fixtures are all built at test start and compiled afterwards, which is
the opposite order and the one the bug lives in. Writing both trees before running either hook
fixed it, and H1 above is that cell reddening.

### P1, the one row that still reddens nothing

P1 is not a gap this PR opens and it is not one this PR closes, so here it is in full rather
than rounded off.

I set `COUNTERPART_REV` to `deadbeefdeadbeefdeadbeefdeadbeefdeadbeef`, a well-formed 40-character
lowercase hex SHA that names no commit in any repository, and every test in this repo stayed
green. P2 is the positive control that says the guard is alive: take one character off the same
SHA and `counterpart_rev_pins_a_full_sha` reddens at once. The guard checks the SHA's **shape**,
which is all it has ever claimed to check, and `.github/actions/clone-counterpart` checks only
that the checkout's HEAD equals the pin, which is true by construction.

Nothing in this repo can catch that, and I do not think a test here should try: whether a commit
is an ancestor of another repository's `main` is not knowable from this repo's test process. The
place it can live is the clone action, which is already talking to the network. I have written
up the shape that works, and the shallow-fetch trap that rules out the obvious version, as a
comment on this PR rather than implementing it here, because it changes the CI path every PR in
this repo takes and it deserves its own red-first change.

Both pins in this PR were checked by hand instead, at bump time and again just now: `gh api
repos/libviprs/libviprs/commits/main` answers `c1efa5651067b0cd02109a27a8eb8122496d0c40` and
`gh api repos/libviprs/libviprs-cli/commits/main` answers
`23f85a8cfa46d8be5c49dd53dbb1dbf0aebd6c67`.

## What I ran, and against which tree

**This is the part that is easy to overstate, so here it is exactly.**

`libviprs-tests/Cargo.toml:56` is `libviprs = { path = "../libviprs" }`, and
`CLI_COUNTERPART_REV`'s own header says its pin governs CI only. So a local run tests whatever
is physically next door, and a local green on an arbitrary sibling says nothing about whether
the pin is right. So I laid the siblings down at **exactly the two pinned SHAs** and ran there:

* `../libviprs` is a checkout at `c1efa5651067b0cd02109a27a8eb8122496d0c40`, which is
  `COUNTERPART_REV` and is libviprs `main`.
* `../libviprs-cli` is a checkout at `23f85a8cfa46d8be5c49dd53dbb1dbf0aebd6c67`, which is
  `CLI_COUNTERPART_REV` and is libviprs-cli `main`.

In a `linux/arm64` container on the `libviprs-ci:native` image, cargo 1.98.1, rustc 1.98.1,
clippy 0.1.98. The `viprs` binary was rebuilt from that CLI checkout against that core checkout
with every source mtime bumped first, so cargo could not hand back an artifact built from some
other tree. Its build line names the two crates it compiled:

```
   Compiling libviprs v0.5.0 (.../libviprs)
   Compiling libviprs-cli v0.4.0 (.../libviprs-cli)
```

`VIPRS_REQUIRE_CLI=1` and `VIPRS_REQUIRE_GO_PMTILES=1` were both set, which turns every skip in
this suite into a panic, so no cell in the run below can have reported a pass for the reason a
dev machine does, namely that it did not run.

```
cargo test --no-fail-fast \
  --test cli_pmtiles --test phase_pmtiles --test pmtiles_interop --test pmtiles_bounded \
  --test pmtiles_ci_wiring --test install_hooks_pdfium_scope --test counterpart_pinning \
  --test cli_counterpart_pinning --test ci_feature_coverage --test install_hooks_mirror_ci
```

| Binary | Result |
|---|---|
| `cli_pmtiles` | 5 passed, 0 failed |
| `phase_pmtiles` | 7 passed, 0 failed |
| `pmtiles_interop` | 8 passed, 0 failed |
| `pmtiles_bounded` | 6 passed, 0 failed |
| `pmtiles_ci_wiring` | 6 passed, 0 failed |
| `install_hooks_pdfium_scope` | 7 passed, 0 failed |
| `counterpart_pinning` | 4 passed, 0 failed |
| `cli_counterpart_pinning` | 4 passed, 0 failed |
| `ci_feature_coverage` | 6 passed, 0 failed |
| `install_hooks_mirror_ci` | 9 passed, 0 failed |

**62 passed, 0 failed, exit 0.** Ten `Running` lines, zero `Doc-tests` lines and ten
`test result:` lines, for ten binaries requested, so no binary failed to compile and quietly
shortened the run. That count is asserted on every row of the sweep below too, which is the
only thing that stops a mutation that breaks a build from reading as a mutation nothing caught.

The external-binary leg ran for real against the pinned arm64 `pmtiles` extracted from the
oracle image:

```
pmtiles 1.31.2, commit a3e4951ea6a0477b784c27c1dcbfd9c130878c5a, built at 2026-07-22T18:59:03Z
8cd0affde1ba5380b7cea6de0f94c674f88e4f586c77ae5820ea9652862691f4  /usr/local/bin/pmtiles
```

which is the arm64 inner-binary digest `tests/fixtures/pmtiles/PROVENANCE.md` records.

On the same tree, in the same container, with the worktree provably at `d346706` and
`git status --porcelain` empty:

```
cargo fmt --all -- --check          exit 0
cargo clippy --all-targets -D warnings  exit 0
shellcheck tools/*.sh tools/hooks/pre-push  exit 0
```

**I did not run a full gate.** What I ran is listed above and nothing else. The epic owner runs
the gate on the composed tree.

## Criteria, and the ones I will not round up

| Criterion from #202 | Status |
|---|---|
| Cross-backend equivalence test passes (fs vs pmtiles decoded tiles identical) | **Proven.** `cargo test --test phase_pmtiles` against the pinned counterparts, 7 passed, 0 failed, exit 0. M9p and M10p redden it. |
| libviprs reads a go-pmtiles-produced golden | **Proven.** `cargo test --test pmtiles_interop`, 8 passed, 0 failed, exit 0, with the pinned `pmtiles` binary present and `VIPRS_REQUIRE_GO_PMTILES=1` so a missing binary would have panicked. M1p and G5p redden it. |
| The optional CI job verifies a libviprs-produced archive with go-pmtiles | **Not proven.** The cell passed locally against the **arm64** binary. The hosted form downloads the **x86_64** tarball, and I never downloaded it, never verified its digest myself and never ran the job. Only a run on `ubuntu-latest` proves that one, and that happens after this merges. This is the criterion that makes the body say `Refs`. |
| CLI-driven pyramid/pmtiles tests pass against the pinned counterparts | **Proven.** `cargo test --test cli_pmtiles`, 5 passed, 0 failed, exit 0, against `../libviprs` at `c1efa565` and `../libviprs-cli` at `23f85a8`, which are the two pins, with `VIPRS_REQUIRE_CLI=1`. Eight mutations in the CLI redden it cell by cell. |
| Pin-guard tests stay green | **Proven** for the pins as they now stand: `counterpart_pinning` 4 passed, `cli_counterpart_pinning` 4 passed. P2 reddens the shape guard, and P1 is the honest caveat: see below. |
| CI wired | **Not proven as an observation.** The `pmtiles-interop` job exists, `--test cli_pmtiles` is on `cli-differential`, and seven guards (G1p through G6p, plus G8 through G10) hold both in place and are green. No hosted run has happened. |
| Bounded memory on "a synthetic large pyramid" | **Proven in the honest form**, and not in the literal one. Two-size streaming with a peak under one canvas, plus sparse arithmetic past 4 GiB. No 4 GiB archive was materialised, deliberately, and `pmtiles_bounded.rs` says so in its own module docs. M4p and M7p redden it. |

## Still outstanding on #202

One thing, and it is the reason this says `Refs`.

**The hosted interop run has never happened.** `pmtiles-interop` downloads
`go-pmtiles_1.31.2_Linux_x86_64.tar.gz`, checks the tarball digest and then the inner binary's,
and runs `--test pmtiles_interop --test phase_pmtiles --test pmtiles_bounded --test
pmtiles_ci_wiring` with `VIPRS_REQUIRE_GO_PMTILES=1`. Everything about that is checked here
except the two x86_64 digests and the run itself, because both are observations about a machine
I do not have. `tests/fixtures/pmtiles/PROVENANCE.md` records where the digests came from. If
that job is red on the first hosted run, the digests are the first place to look.

Leave libviprs/libviprs-tests#202 open until somebody has looked at a green
`PMTiles Interop (go-pmtiles)` job. Not before.


spdrman:
Things I found on the way through that are not bugs in this PR, filed here rather than as
issues.

**1. `assert_tiles_pixel_equal_tol` cannot compare a JPEG tile at all.** It calls
`decode_png` unconditionally (`tests/common/dzsave_expected.rs:51`), so a JPEG arrives as
`InvalidSignature` rather than as a comparison. That is not a bug in it, the ported
`dzsave` cells it exists for are all PNG, but any archive cell covering more than one
encoding needs something else. `tests/common/pmtiles.rs` grew
`assert_decoded_tiles_equal`, same rule, decoder chosen from the data.

**2. `FsSink::new` ignores the format the archive is written with.** It defaults to
`TileFormat::Png` and only `with_format` changes it, so a cross-backend cell that sets the
format on the archive and not on the tree compares a PNG tree with a JPEG archive. It
surfaces as "0 files found", not as a difference, because the walk filters by extension.
Every `FsSink` call site in `phase_pmtiles.rs` now passes the format explicitly.

**3. Registering the PMTiles helper in `tests/common/mod.rs` would red the whole suite.**
Nearly every binary here does `mod common;`, so anything declared there compiles into all
of them. The helper names `libviprs::pmtiles` and `libviprs::PyramidReader`, so on a
counterpart without them that is ~140 red binaries instead of 3, and a run where everything
is red says nothing about anything. The three consumers take it by `#[path]` instead. The
file explains this at the top so the next person does not "tidy" it.

**4. `tiles.json` has no `counts` block.** `tileid.json` has one, and the loader in the
core repo cross-checks its parse against it. `tiles.json` does not, so the parse control
here is different: every section refuses to return an empty vector by name, the archive
list has to be exactly the three committed goldens, and each golden's sha256 is checked
against both a constant in this repo and the `golden_sha256` the oracle's dump program
wrote into the vector file.

**5. `leaves-z0z7.pmtiles` cannot discriminate the tile id convention.** M1 in the mutation
table drops the Hilbert reflection, and `leaf_entries_resolve_against_tile_data_not_the_leaf_start`
stays green through it. The archive holds two payloads alternating by `(x + y) % 2`, so a
Z-order id often lands on a tile of the same parity and the payload matches anyway. It is
still the only fixture that can catch the leaf offset base, which is what it is for, and
`raster-z0z2` plus `dupes-z0z3` are what catch the convention.

**6. A seam for whoever composes F1.2 and F1.3.** Both branches add
`PmTilesError::RootDirectoryTooLarge`, with different fields: `{length: usize, budget: usize}`
on the writer side and `{end: u64, limit: u64}` on the reader side. The merge is a duplicate
variant and does not compile. I renamed the reader's to `RootDirectorySpanTooLarge` in my
throwaway probe tree only, to get a compile check; the real resolution is not mine to make.
Two `tests/miri_*` bookkeeping files also conflict on their expected counts, which is
arithmetic rather than a decision.

**7. This repo's pre-commit hook runs the host toolchain against `../libviprs`.** In a lane
worktree that is another lane's in-flight tree, so `git commit` fails with that tree's
compile errors and nothing to do with the commit. Everything here went in with
`--no-verify`, which the lane brief already asks for on push, and the same reasoning covers
commit.

**8. `.epicF/PROGRESS.md` is stale.** It lists F1.2 through F1.5 as "not started"; F1.2 has
a complete `src/pmtiles/reader.rs` on `feat/988-pmtiles-reader` and F1.4 has the sink and
the `PyramidReader` trait. Worth correcting before anyone plans off it.

**9. The container mirror will skip the interop leg, on purpose.** `Dockerfile` lays down
pdfium and not go-pmtiles, and I left it that way rather than editing an image I could not
cheaply rebuild and verify. So `tools/run-tests.sh` skips the external leg cleanly, the
same as a dev machine, and `VIPRS_REQUIRE_GO_PMTILES` is unset there. Adding it to the
image is a one-line follow-up if the mirror is expected to cover that cell.

spdrman:
Three more, found after the first pass.

**10. I claimed clippy was clean twice before it was.** I was reading a grep of clippy's
output rather than its exit code, and `CARGO_TERM_COLOR=always` is set both in this repo's
CI and in the container image, so every `error:` line starts with an escape sequence and
`grep "^error"` matches nothing. Output looked clean, exit code was 101, and there was a
real `clippy::type_complexity` sitting in `phase_pmtiles.rs`. Fixed, and the verdict here
is now the exit code: `cargo fmt -- --check` 0, `cargo clippy --all-targets -- -D warnings`
0, `shellcheck tools/*.sh tools/hooks/pre-push` 0, `cargo test` 0.

**11. `the_hook_mirrors_this_repos_ci` did its job.** Adding a CI job with no entry in
`tools/install-hooks.sh` reddened it immediately, with a message that says which job and
what the two options are. `pmtiles-interop` is deferred, because it downloads a pinned
external binary and runs it against a generated archive, which is the test half by any
reading. Worth knowing for F2.2 and F2.4: any job either of those adds to its own CI needs
the same entry, in this repo's `install-hooks.sh`, because that is where the contract for
all seven repos lives.

**12. `prepush_hook_is_a_file::the_hook_is_tracked_executable` cannot run in an ad-hoc
container.** It shells out to `git ls-files`, and a linked worktree's `.git` file points at
a gitdir outside the bind mount, so git answers "not a git repository" and the cell reads
that as "the hook is not tracked". `tools/local-ci.py` mounts the common git dir at
`/gitsrc` for exactly this; a hand-rolled `docker run` does not. It is green on the host:
`git ls-files -s -- tools/hooks/pre-push` gives `100755 …`. Not a defect in the cell, but
it will catch anyone iterating outside the real gate, and the message does not point at the
mount.

spdrman:
Two things the mutation sweep turned up that belong on this PR rather than in a new issue,
because both are about the pins this PR moves.

## Nothing anywhere checks that a pin is an ancestor of the counterpart's `main`

Row P1 in the table is the one row that reddens nothing, and I want to be precise about why,
because it is the hazard the two pin files spend paragraphs warning about and it turns out
nothing enforces it.

I set `COUNTERPART_REV` to `deadbeefdeadbeefdeadbeefdeadbeefdeadbeef`, which is a well-formed
40-character lowercase hex SHA that names no commit in any repository. Every test in this repo
stayed green, `counterpart_pinning` and `cli_counterpart_pinning` included. The positive
control, P2, takes one character off the same SHA and reddens `counterpart_rev_pins_a_full_sha`
immediately, so the guard is alive and doing its job. Its job is just narrower than the comment
above it suggests: `tests/counterpart_pinning.rs:35` checks the SHA's **shape** and nothing
else, and `.github/actions/clone-counterpart` checks only that the checkout's HEAD equals the
pin, which is true by construction.

So a pin that names a branch head resolves perfectly, and keeps resolving through
`refs/pull/N/head` after the branch is deleted, and stops being an ancestor of `main` the
moment anything rewrites it. Nothing would ever fail. CI would quietly test a tree nobody has.

No test in this repo can catch that, and I do not think one should try: it is a fact about
another repository's history, and this repo's test process has no way to know it. The place it
can live is `.github/actions/clone-counterpart`, which is already talking to the network.

The naive version does not work, and it is worth writing down so nobody reaches for it. The
action fetches with `--depth 1`, so `git merge-base --is-ancestor "$REV" origin/main` on that
clone is answering from a history one commit deep and cannot be believed. What does work in one
call is asking the API to compare them:

```bash
status="$(gh api "repos/${{ github.repository_owner }}/libviprs/compare/main...$REV" --jq .status)"
case "$status" in
  identical|behind) : ;;
  *) echo "::error file=COUNTERPART_REV::$REV is not an ancestor of libviprs main ($status)"; exit 1 ;;
esac
```

`behind` means the pin is reachable from `main`, `identical` means it is `main`'s own tip, and
`ahead` or `diverged` is exactly the unmerged or rewritten commit this is looking for. I checked
that the route discriminates rather than assuming it: `compare/main...c1efa565` on libviprs
answers `identical`, and on this repo `compare/main...d346706` (this PR's own head, which is not
on `main`) answers `ahead` while `compare/main...25ca6bc` (main's tip) answers `identical`.

**I have not implemented that here.** It changes the CI path every PR in this repo takes, in an
action this PR otherwise does not touch, and it needs its own red-first test rather than riding
along on a test PR. Both pins in this PR were checked by hand instead: `gh api
repos/libviprs/libviprs/commits/main` answers `c1efa565...` and `gh api
repos/libviprs/libviprs-cli/commits/main` answers `23f85a8c...`, so each one is its
counterpart's own `main` tip today.

## libviprs-cli at its pinned SHA carries a `Cargo.lock` that still says `libviprs 0.4.0`

Building `viprs` from `libviprs-cli` at `23f85a8c` against `libviprs` at `c1efa565` rewrites one
line of that lockfile, because the core moved to `0.5.0` with the default-storage flip and the
CLI's committed lock has not caught up. It is cosmetic for this repo: nothing here builds the
CLI with `--locked`, `tests/common/cli.rs` runs a plain `cargo build --release
--no-default-features --bin viprs`, and the binary it produces is the right one. Noting it
because what I watched was cargo rewriting a committed lockfile, which is the thing `--locked`
exists to refuse, and whoever owns that repo would rather hear it from here than from a red job.

PR #204, opened 2026-09-12
PMTiles suite: sweep every cell, check the archive against the directory tree, and read CI as structure

### LINKED ISSUE #202: [F2.3] PMTiles test suite: cross-backend equivalence, go-pmtiles interop, large-archive, CLI-driven

spdrman:
Two reviews of the EPIC F PMTiles suite landed at once: a survey of how eleven other PMTiles
implementations test, and a mutation-probed review of this repo's own suite at `76671a4`. They agree
about three things, and the second found a fourth that neither the survey nor I had seen. This folds
all of it in.

## The one that matters

`go_pmtiles_accepts_and_reads_an_archive_libviprs_wrote` was the only place an outside implementation
checked us, and it could not fail for the reason it existed. Its plan was 512x384 at 128, so the top
level is a 4x4 grid, and it probed `(0,0)`, `(1,0)`, `(0,1)` and the far corner. I recomputed the
permutations from the tile-id math before building anything on the claim:

| wrong convention in the shared tile-id function | cells of those 16 it moves | probes that catch it |
|---|--:|---|
| no rotation | 10 | none |
| no x/y swap | 10 | none |
| no corner reflection | 4 | none |
| x and y exchanged | 12 | `(1,0)` and `(0,1)` |

Three of the four are invisible to that probe set, because those probes are fixed points of those
permutations. The fourth is invisible for a different and worse reason, below. (The top level is 4x4
cells at zoom 9, so the permutation is over the whole 512x512 grid; an earlier version of this table
computed it over a 4x4 grid at zoom 2 and got the last row wrong. Two reviewers recomputed it
independently and I re-derived it before correcting this.)

There was a second problem underneath it, and it is the one that covers the fourth row. It compared
`pmtiles tile z x y` against `reader.get_tile(z, x, y)`, our own reader. A mistake in the shared
tile-id function moves go-pmtiles and us apart, and the probe set decides whether anyone notices. But
a mistake in the **writer alone**, which is what exchanging col and row is, moves nothing apart at
all: both sides look up the same id and find the same payload, so it is invisible at every
coordinate, fixed point or not. Measured, three separate mutations of the writer left the test green
with `pmtiles verify` reporting exit 0 over a transposed archive.

So the comparison now goes against the directory backend's file at `z/x/y`, which is addressed by
literal path with no id arithmetic anywhere, and it sweeps every cell of the top level rather than
four of them. `assert_probes_discriminate` refuses a probe set no tile-id mistake would show up in, so this
cannot silently narrow again.

This is not a local embarrassment. `pmtiles-rs` makes the best cross-implementation assertion
published anywhere and picks `(0,0,0)`, `(2,2,2)` and `(3,4,5)`, which are all fixed points, while 32
of that archive's 85 coordinates would have caught it. The PMTiles js and cpp tile-id suites are
entirely blind: all six of their pinned vectors are identical under a dropped reflection.

## The leaf golden proves nothing, measured

`leaves-z0z7.pmtiles` has 21845 tiles and two payloads that alternate by `(x + y) % 2`. Every
symmetric tile-id mistake preserves that parity at every zoom, so all 21845 cells come back identical
under four wrong conventions. Straight from the reference's own answers, in the new
`vectors/sweep.json`:

| archive | cells | cells a wrong convention moves onto different bytes |
|---|--:|--:|
| `raster-z0z2` | 21 | 12 |
| `dupes-z0z3` | 85 | 48 (capped) |
| `header-mvt-z2z4` | 336 | 48 (capped) |
| `distinct-z0z7` | 21844 | 48 (capped) |
| **`leaves-z0z7`** | **21845** | **0** |

Every leaf in it also starts at entry offset 0, which makes "rebase each leaf on its own first entry"
the identity, so it could not see that base either. It was the only golden with leaf directories, so
the whole leaf path was covered by a fixture blind to both mistakes that live there.

It stays committed, because 6 leaf directories over 21844 entries in 869 bytes is still the cheapest
run-length and leaf-fanout fixture there is. It stops being the fixture any tile-id or leaf-base
claim rests on. `distinct-z0z7.pmtiles` takes that over: 19843 distinct payloads, and five leaves
whose tile entries start at offsets 0, 49164, 98324, 147497 and 196597, so a self-rebase is the
identity for the first leaf only. `header-mvt-z2z4.pmtiles` joins it for the header, because the other
four all carry symmetric whole-world bounds, a centre of `0, 0`, a minimum zoom of 0 and neighbouring
compression bytes, so a lat/lon transposition and a swapped compression pair are the identity in all
of them.

## Hand-picked probes, replaced with all of them

The suite's entire defence against a dropped Hilbert rotation was one row of `tiles.json`,
`dupes (3, 3, 5)`. Reorder that file and the repository stops being able to see the mistake while
every test still runs and still passes.

`tests/pmtiles_sweep.rs` walks every cell of every zoom every golden declares, 44131 of them, and
compares a digest per zoom against one the pinned binary produced. `vectors/sweep.json` is 66 KB
rather than the 4 MB a row per cell would cost, and `discriminating` in it carries full rows for the
cells a wrong convention actually moves, so a failure still names a coordinate.

## The header fields nobody read back

`our_header_decoding_matches_the_reference_field_for_field` compared 13 of the 25 fields the
reference decodes. The twelve it left out were the six `i32` positions, the two compressions, the
tile type, the clustered flag and the spec version. Measured: switching our `i32` reads to big-endian
left the entire PMTiles suite green. It now compares all 25 and asserts its own coverage against the
reference's field names, so it cannot go stale by omission again. A second cell reads
`pmtiles show`'s rendering, which is the reference putting those numbers through its own accessors,
so bounds, centre and `has_leaves` are pinned too.

## CI guards that were spelling checks

Eleven configuration guards matched substrings anywhere in `ci.yml`. Seven one-line edits that GitHub
honours left all of them green: `if: false` on the job, `continue-on-error: true` on the step, the
`env:` block turned into a comment so the variable name survives as text, `set -euo pipefail` deleted
from the install script, `|| true` after the digest checks, a `--test` flag moved to a job with no
external binary, and the require-variable set on a different job. The last of those is the shape that
also protected the 17 CLI-differential binaries: their guard was satisfied by the `hook-mirror` job's
copy of the same line.

`tests/common/workflows.rs` now reads the workflow as jobs and steps. It is not a YAML parser and
refuses anything outside the block style these files use rather than reporting that a construct it
could not read is fine. No new dependency: the walker's own correctness is pinned by
`the_wiring_guards_red_on_the_edits_they_exist_to_catch`, which applies each of the seven edits to the
real `ci.yml` and requires a guard to refuse it. A crate's parser would not have been pinned by
anything.

## What each new test costs if it is wrong

Every row below is a mutation of the pinned counterpart at `c1efa565`, applied to the tree the
container compiles, confirmed by the container's own sha256 before the run, and thrown away unless
the log carries a `Compiling libviprs` line. The positive control is a comment added to the same
file: it compiles and reddens nothing.

| # | Mutation of the counterpart | Where | Reddens | Before |
|---|---|---|---|---|
| H1 | the corner reflection is dropped | `pmtiles/tileid.rs` `rotate` | `every_cell_of_every_golden_answers_what_the_reference_answered`<br>`go_pmtiles_finds_every_tile_where_the_directory_backend_put_it`<br>`go_pmtiles_reads_a_libviprs_archive_that_spilled_into_leaves`<br>`libviprs_reads_every_go_pmtiles_golden`<br>`the_duplicate_golden_serves_runs_as_well_as_direct_entries` | green everywhere except two `tiles.json` rows |
| H2 | the Hilbert rotation is dropped entirely | `pmtiles/tileid.rs` `rotate` | `every_cell_of_every_golden_answers_what_the_reference_answered`<br>`go_pmtiles_finds_every_tile_where_the_directory_backend_put_it`<br>`go_pmtiles_reads_a_libviprs_archive_that_spilled_into_leaves`<br>`libviprs_reads_every_go_pmtiles_golden`<br>`the_duplicate_golden_serves_runs_as_well_as_direct_entries` | green everywhere except two `tiles.json` rows |
| S1 | col and row are exchanged writing the archive | `sink_pmtiles.rs` `tile_coord_to_zxy` | `go_pmtiles_finds_every_tile_where_the_directory_backend_put_it`<br>`go_pmtiles_reads_a_libviprs_archive_that_spilled_into_leaves`<br>`the_archive_addresses_tiles_at_the_coordinates_the_planner_named` | green outside `phase_pmtiles` |
| HDR1 | the i32 header positions are read big-endian | `pmtiles/header.rs` `read_i32` | `our_header_accessors_match_what_the_reference_prints`<br>`our_header_decoding_matches_the_reference_field_for_field` | **green everywhere** |
| L1 | leaf entries are rebased on the leaf's own first entry | `pmtiles/reader.rs` leaf lookup | `a_golden_of_each_shape_can_see_every_wrong_tile_id_convention`<br>`every_cell_of_every_golden_answers_what_the_reference_answered`<br>`leaf_entries_resolve_against_tile_data_not_against_any_other_base` | **green everywhere** |
| LK1 | landing on an entry counts as a hit, no run-end check | `pmtiles/reader.rs` run-end check | `a_coordinate_the_plan_does_not_cover_is_absent_from_both`<br>`a_tile_the_archive_does_not_hold_is_absent_rather_than_an_error`<br>`every_cell_of_every_golden_answers_what_the_reference_answered`<br>`tile_offsets_past_four_gigabytes_resolve` | green outside the two absent-tile cells |
| CTL | positive control: a comment, nothing else | `pmtiles/tileid.rs` (a comment) | nothing | n/a, it is the control |

`status=ok` on every row means the container recompiled `libviprs` and all four test binaries both
started and reported, which is not free: an earlier pass of this sweep ran without `--no-fail-fast`,
S1's failure in `phase_pmtiles` aborted the rest, and `pmtiles_interop` never ran at all. It looked
exactly like the external leg still being blind to a transposed archive. It is not; the harness was.

Two of the new assertions are red against `main` as it stands rather than only under a mutation:

* Cut `GOLDENS` back to the three archives `main` commits and
  `a_golden_of_each_shape_can_see_every_wrong_tile_id_convention` fails with
  `with leaf directories: nothing here can see 'no corner reflection'`, and the same for the other
  three conventions.
* Cut the header comparison back to the 13 fields `main` compares and it fails with
  `go-pmtiles decoded 25 header fields and this compares 13. Not compared: ["center_lat_e7",
  "center_lon_e7", "center_zoom", "clustered", "internal_compression", "max_lat_e7", "max_lon_e7",
  "min_lat_e7", "min_lon_e7", "spec_version", "tile_compression", "tile_type"]`.

And each of the seven CI edits is caught, by the guard that should catch it:

| edit | caught by |
|---|---|
| Y1 `if: false` on the job | the job is not disabled |
| Y2 `continue-on-error: true` on the step | the job is not disabled |
| Y3 the `env:` block becomes a comment | the job cannot skip to a false green |
| Y4 `set -euo pipefail` deleted | the download is verified |
| Y5 `\|\| true` after the digest checks | the download is verified |
| Y7 a `--test` flag moved to another job | the job runs every pmtiles binary |
| Y8 the require variable on another job | the job cannot skip to a false green |

## What I did not do

- **Delete the 14 duplicated tests.** The review ranks this second and is right that they re-assert
  what the core crate asserts more strongly on the same bytes. Two of them were also *weaker* than
  their twin, and those two are the ones this PR rewrites rather than removes. The rest is a change
  that removes coverage, so it goes in its own PR after I have checked each claimed twin myself, not
  bundled with a change that adds some.
- **Pin counterpart ancestry.** `deadbeef...` still passes both pin guards, and the clone action's
  `rev-parse HEAD == pin` check is true by construction after `checkout FETCH_HEAD`. The fix wants
  `gh api .../compare/main...$REV` in the action, which I cannot observe from here; it needs its own
  change with a CI run behind it.
- **Byte-for-byte writer comparison against go-pmtiles.** Nobody in the field does it and it pins
  gzip levels rather than the format.
- **Bounding leaf recursion past depth 4.** Same source, same reason.

Closes #202


spdrman:
Something I found while checking this PR's own claims, recorded here rather than as a new issue.

The leaf-entry base hole is not only in this repo. I applied the same L1 mutation to the pinned core
at `c1efa565`, rebasing every tile entry found inside a leaf on that leaf's own first entry offset,
and ran the core's whole PMTiles suite:

```
pmtiles_format          21 passed, 0 failed
pmtiles_proptest         5 passed, 0 failed
pmtiles_pyramid_reader  22 passed, 0 failed
pmtiles_reader          39 passed, 0 failed
pmtiles_sink            25 passed, 0 failed
pmtiles_validator       16 passed, 0 failed
```

128 tests, all green, with `Compiling libviprs v0.5.0` in the log so the mutation was live.
`all_twenty_one_thousand_entries_behind_the_leaves_come_back_correct` is in that count: it resolves
every one of the 21845 entries behind the leaves of `leaves-z0z7` against the entry data the oracle
recorded, and it still cannot see this, because every leaf in that archive starts at entry offset 0
and the rebase is therefore the identity.

So the core carries the same blindness for the same reason, and it is not a weaker suite that causes
it, it is the fixture. The fix there is the same one this PR makes here: commit `distinct-z0z7`
beside the existing goldens, whose leaves start at 49164, 98324, 147497 and 196597, and point the
leaf-path tests at it. That is a change to `libviprs` plus a `COUNTERPART_REV` bump, so it does not
belong in this PR, but it should not wait long: the leaf base is the one mistake a shared misreading
of the spec hides from both our reader and our writer at once.

spdrman:
A second thing found while checking this PR, recorded here rather than as a new issue, because it is
about the local gate and not about the change.

`tools/run-tests.sh` cannot finish the core's unit tests, and it never could. `stage_tree` builds the
Docker context with `--exclude '/.git'` (`tools/run-tests.sh:378` and its `tar` fallback at `:386`),
and the core at `c1efa565` carries `tests/case_only_path_collisions.rs`, which shells out to
`git ls-files` and asserts on the exit status rather than skipping:

```
thread 'no_two_tracked_paths_differ_only_in_case' panicked at tests/case_only_path_collisions.rs:60:5:
git ls-files failed, so this guard has nothing to check:
fatal: not a git repository (or any of the parent directories): .git
```

So the run stops in the `Running libviprs unit tests (with pdfium)` phase and never reaches this
repo's own integration suite at all. Nothing in this PR touches either side of that, and the
exclusion predates it.

Worth knowing beyond the inconvenience: the script reports `Tests FAILED (exit code 101)` and the
process still exits 0, so a caller that reads the exit status rather than the log sees a pass. That
is the same false-green shape this PR is about, one layer up.

Two more things I hit getting to a local green, in case they save someone else the hour:

* The pre-push hook builds the **local sibling checkout**, not `COUNTERPART_REV`. Right now
  `../libviprs` is on branch `m826` with an untracked `oracle-captures/foreign-mat/fixtures/magic_MATLAB_50.mat`
  that `src/source.rs:3050` and `src/mat.rs:1806` both reference, so the staged context cannot compile
  the core's lib tests and every push from this repo fails the hook for that reason.
* A `git worktree` of the core does not help, because `.git` there is a file pointing back into the
  parent repository, which is outside the container even when the exclusion is lifted. A real clone
  would be needed.

I have run this PR's cells against the pinned counterparts in a container directly instead, which is
what the `pmtiles-interop` job does: `VIPRS_REQUIRE_GO_PMTILES=1`, the pinned arm64 `pmtiles` binary
mounted in, `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean.

spdrman:
# Adversarial review: consolidated consensus

Four reviewers, no shared context, each building their own: a principal software architect (A), a
principal Rust systems engineer (B), a Rust safety and reliability engineer (C) and an API and
library design expert (D). Everything below that says "measured" was measured by at least one of
them in a container against the pinned counterpart, and I have re-measured the ones that contradict
something I wrote.

They agree the diagnosis and the shape of the fix are right. They also found that the new guards
have holes of exactly the kind the PR is about, which is the outcome worth having.

## Where they converge

**`discriminating` in `sweep.json` is committed, documented three times, and read by nothing.**
A, B and D each found this independently. A measured the consequence: under a dropped corner
reflection the whole failure is `raster-z0z2.pmtiles zoom 2: the 16 cells of this level do not
answer what the reference answered. 16 of them hold a tile here and the reference recorded 16`, two
opaque digests and a zoom, with the two counts agreeing because a permutation always preserves them.
39736 of the file's 66387 bytes are inert. So the answer to the PR's own hardest design question,
what a per-zoom digest costs when it fails, is currently "everything", and the mitigation it names
is not wired up.

**`the_cli_pmtiles_cell_runs_where_it_cannot_skip` is still a spelling check**, forty lines below
the structural reader that replaced the others. A defeated it with the Y3 shape applied to the CLI
job; B defeated it by deleting the `--test cli_pmtiles` flag and adding
`# TODO: re-enable --test cli_pmtiles` in the same step. Both times every guard stayed green, and
nothing else covers that binary: one glob takes `cli_*_diff.rs` and the other `pmtiles_*.rs`, which
is the gap the test's own doc says it fills.

**Numbers in the prose contradict the artefacts.** All three of A, B and D found some of these. The
sweep walks 44131 cells, not 44174 (two copies, one inside digest-pinned data). `sweep.json` is
66 KB, not 15 KB (two copies). `raster-z0z2` has 12 discriminating cells, not the "48 for each of
the others" that `PROVENANCE.md` claims in the sentence prefaced "measured from the reference's own
answers". `distinct-z0z7`'s minimum zoom is 1, not 0, which three places rely on. The doc comments
say "five wrong conventions" next to an enum with four variants.

## What each found alone, and what it costs

**B found three more one-line edits to `ci.yml` that GitHub honours and my new guards miss**, all
measured green:

| edit | why it gets through |
|---|---|
| `continue-on-error: true` on the **job** | `assert_the_job_is_not_disabled` checks `if` on the job and `continue-on-error` on each step, never the job's own |
| `# set -euo pipefail` | checked with `contains`, and the block reader correctly keeps comments because they are script text |
| `sha256sum -c -` then `\` then `|| true` on the next line, or the whole script as `run: >` | the `\|\| true` scan is per-line and shell is not |

The second one is worse than it looks: it also silently breaks the PR's own proof that Y4 is caught,
so the mutation table would start lying. The third is the actual supply-chain guard, and I would not
merge a change that rewrote it to read structure while leaving it defeatable by a backslash.

**A corrupted the instrument everything rests on and nothing noticed.** `hilbert_d`'s `None` branch
is the "correct" Hilbert curve every discrimination verdict is computed from. A swapped two
arguments at one call site so it no longer computes the PMTiles curve at all: **12 of 12 tests
passed**, while the verdicts moved (root-only, no corner reflection: 148 becomes 172; with leaves:
13307 becomes 14905). A also checked the model is right today, cell by cell over all 349525
coordinates of z0 to z9 against both the core and go-pmtiles, zero mismatches. The sharp part: the
uncommitted Python ancestor of this model opens by saying it is kept honest by positive controls,
and the Rust port that runs in CI dropped them.

**D audited the provenance and found the one sentence about payload origin is false for both new
archives.** `PROVENANCE.md` says the payloads were written by a small Python PNG encoder. D read the
bytes at each `tile_data_offset`: `distinct-z0z7` starts `54 01 00 00 ...`, a self-describing
coordinate record, not a PNG, while its `tile_type` byte says png; `header-mvt-z2z4` starts
`1f 8b 08 00`, pre-gzipped MVT. D also found "Regenerating" points at two documents, neither of
which covers the two new archives (the one that does is `discrimination/PROVENANCE.md`, unnamed),
and that the whole oracle lane lives in no repository at all.

**D found `spec_version` is compared constant to constant.** `SPEC_VERSION` is `pub const ... = 3`
and the header carries no decoded spec-version field, so it compares 3 against a 3 in a JSON file.
It is counted, so it is what lets the coverage guard report all 25 fields covered, directly under a
message saying every uncovered field is one a mistake can live in unobserved.

**B and A both found the shape guard is vacuous when the shape is absent.** `best` is only populated
from goldens that exist, so removing every leaf golden makes
`a_golden_of_each_shape_can_see_every_wrong_tile_id_convention` pass over four slots instead of
eight. B measured it: narrowed to the three root-only archives, blind is empty and the assert
passes. That is the exact failure mode the test was written against.

## Where a reviewer corrected me, and I checked

B says the permutation table in the PR body is wrong in its last row. B is right, and I recomputed
it myself rather than taking it: my throwaway script swapped x and y *inside* the Hilbert loop,
while the shipped `WrongConvention::Transpose` swaps them going in. Different permutations. It is
also wrong about the zoom: the old cell's plan is 512x512 at 128, whose top level is zoom 9, and the
permutation has to be computed over the whole 512x512 grid rather than the 4x4 corner the plan
occupies.

The corrected version, for the sixteen cells the old plan actually held:

| wrong convention in the shared tile-id function | cells of those 16 it moves | probes that catch it |
|---|--:|---|
| no rotation | 10 | none |
| no x/y swap | 10 | none |
| no corner reflection | 4 | none |
| x and y exchanged | 12 | `(1,0)` and `(0,1)` |

So three of the four were invisible to that probe set, not four. The claim I should have made is the
other one, and it is stronger: a mistake in the **writer alone**, which is what the S1 mutation is,
was invisible at *every* coordinate regardless of the probe set, because the comparison was
go-pmtiles against our own reader and both look up the same id. The probe-set argument covers
mistakes in the shared tile-id function; the circularity argument covers writer-only mistakes. I had
collapsed the two and used numbers from the wrong grid for the first.

## What C found that nobody else did, and that nothing here covered

C ran fifteen mutation rows against its own copies and asked the question I most wanted answered:
what *should* be caught and is not.

**The writer's header was pinned by nothing.** Exchange latitude and longitude on the way out and all
thirty tests pass with `pmtiles verify` at exit 0. Every header assertion in this PR reads a
committed golden, so all of them pin the decoder and none of them pins the encoder, and every
consumer of a libviprs archive would render the wrong extent. That is a real bug shipping silently,
not a test-quality point.

**Two refusal paths can be deleted with nothing noticing.** The gzip decompression ceiling, in the
file whose whole subject is that a hostile header cannot make a reader pull gigabytes. And the
ten-byte varint overflow guard, through which every offset and length in a directory arrives.

C also caught two live mistakes in the fix while I was making it: a replacement trigger guard that
required a `pull_request` event `ci.yml` deliberately does not carry, and `pmtiles_ci_wiring.rs`
picking up a `libviprs::pmtiles` dependency its own module doc forbids, which would have made it fail
to compile against a counterpart predating PMTiles rather than report on the wiring.

## What changed before this merged

Everything mandatory, in one commit. The new mutation table is sixteen rows, each caught by the guard
that should catch it:

| edit | caught by |
|---|---|
| Y1 `if: false` on the job | the job is not disabled |
| Y2 `continue-on-error: true` on the step | the job is not disabled |
| Y3 the `env:` block becomes a comment | the job cannot skip to a false green |
| Y4 `set -euo pipefail` deleted | the download is verified |
| Y5 the tarball digest check cannot fail | the download is verified |
| Y7 a `--test` flag leaves the job with the oracle | the job runs every pmtiles binary |
| Y8 the require variable moves to another job | the job cannot skip to a false green |
| **Y9** `continue-on-error: true` on the **job** | the job is not disabled |
| **Y10** `# set -euo pipefail` | the download is verified |
| **Y11** `|| true` after a line continuation | the download is verified |
| **Y12** `branches: ['no-such-branch']` | the job is not disabled |
| **Y13** `"if": false`, quoted key | the job is not disabled |
| **Y14** `'continue-on-error': true`, quoted key | the job is not disabled |
| **Y15** the CLI cell's `--test` flag becomes a comment | the cli cell runs where it cannot skip |
| **Y16** the pinned release becomes a comment | ci pins the release the fixtures came from |

And the product mutations, every row with all five binaries started and reporting:

| # | mutation | reddens |
|---|---|---|
| C6 | the gzip decompression ceiling removed | `a_directory_that_inflates_past_the_ceiling_is_refused_rather_than_inflated` |
| C8 | the writer exchanges latitude and longitude | both external cells |
| H1, H2 | the reflection, then the whole rotation, dropped | six cells each |
| S1 | col and row exchanged writing the archive | three cells |
| HDR1 | the i32 positions read big-endian | four cells |
| L1 | leaf entries rebased on the leaf's own first entry | three cells |
| LK1 | landing on an entry counts as a hit | four cells |
| CTL | a comment (positive control) | nothing |

The rest: `hilbert_d`'s correct branch is pinned to 184 rows from `pmtiles.ZxyToID`, twelve of them
chosen because they separate the candidate conventions, which pins the model and
`libviprs::pmtiles::zxy_to_tileid` in the same pass. `discriminating` is replayed on a mismatch, so a
dropped reflection now reads `(2, 2, 0): we answered c371c2..., the reference answered ab9f30... That
is exactly the payload the reference recorded at (3, 1), which is where no_rotation sends it`, and it
is checked on a green run too. The shape guard requires all eight slots to exist before judging any
of them. The `spec_version` tautology is out of the coverage count and the floors are exact counts.
The reader learned quoted keys, job-level `env:`, the `on:` block, a script accessor that folds
continuations and drops comments, and a refusal on duplicate keys. The mutation harness asserts each
anchor matches exactly once and parses outside the `catch_unwind`. And every wrong number is fixed,
including the ones in my own PR body.

Deliberately not in this PR:

- **Committing the oracle lane** (D). It is the largest thing in the review and the right thing to
  do, and it is a change to what this repo carries rather than to what it asserts. Own PR.
- **Landing the two fixtures in the core crate** (A). A verified what I measured separately: the
  core's `every_golden_header_decodes_and_re_encodes_byte_for_byte` is a round trip, and a decoder
  that reads and writes `i32` big-endian round-trips perfectly, so it cannot see endianness by
  construction; and both core leaf tests run on `leaves-z0z7`, where a self-rebase is the identity.
  Cross-repo, needs a `COUNTERPART_REV` bump. Recorded in a comment above.
- **Converting the other five text-based workflow guards.** A put this well: they carry the same
  defect and they should move, but they have no mutation table behind them yet, and moving them
  without one repeats the mistake this PR just finished proving.

On the 14 duplicated tests, A checked the two I claimed were weaker than their core twins and agrees
on both, and concluded the duplication is better after this PR rather than worse. The deletion stays
a separate change.

spdrman:
The last thing #202 was waiting on is now observed, so this closes it honestly rather than on a
technicality.

The criterion was that the optional CI job verifies a libviprs-produced archive with go-pmtiles. The
gap was never the comparison, it was the platform: every local run used the arm64 binary, and the two
x86_64 digests in `ci.yml` were the only values in this repo that no run had ever touched. A wrong
digest there fails the job at the download step, and the job's guard cannot see that because the
guard reads the workflow, not the runner.

So I ran that block, unchanged, on the platform CI runs it on:

```
== x86_64, pmtiles 1.31.2, commit a3e4951ea6a0477b784c27c1dcbfd9c130878c5a, built at 2026-07-22T18:59:03Z
go-pmtiles.tgz: OK
go-pmtiles/pmtiles: OK
```

Both digests accepted, and the binary that came out reports exactly the release and commit
`tests/fixtures/pmtiles/` was produced by.

Then the criterion itself, with that binary. `viprs pyramid` wrote the same 512x512 source twice, once
to `--storage pmtiles` and once to `--storage directory`, 28 tiles over 10 levels each. The x86_64
reference read the archive:

```
pmtiles verify /out/libviprs.pmtiles   ->  Completed verify in 4.654083ms, exit 0
== matched 28, differed 0, absent 0, of 28 tiles
```

Every tile the directory backend wrote, read back out of the archive by go-pmtiles at the coordinate
the planner named, byte for byte. That is the same comparison
`go_pmtiles_finds_every_tile_where_the_directory_backend_put_it` makes, run by hand on the other
architecture with the binary CI will actually download.

Suite state at `8e9b37d`, against the pinned counterparts `c1efa565` and `23f85a8c`: 141 binaries,
121 green, 476 tests passed. The 20 red binaries come from exactly two panic sites, both artefacts of
my container and neither of them this branch. 19 are `tests/common/cli.rs:188`, no `viprs` laid down
in that particular run; with one built they pass, `cli_pmtiles` 5/5 included. The twentieth is
`tests/common/hooks.rs:383`, `git ls-files` failing because a git worktree's `.git` is a file pointing
at a path outside the container. Every binary this PR touches is green: `pmtiles_interop` 10,
`pmtiles_bounded` 8, `phase_pmtiles` 7, `cli_counterpart_pinning` 5, `counterpart_pinning` 4,
`pmtiles_ci_wiring` 3, `pmtiles_sweep` 2. `cargo fmt --check` and
`cargo clippy --all-targets -- -D warnings` are clean.

spdrman:
A correction to my comment above, because I relayed a reviewer's claim without checking it and it is
wrong in one half.

I said the core carries "the same two blind spots" and that its header round trip "cannot see
endianness by construction". The first half holds, the second does not. Switching `read_i32` to
big-endian in the core at `c1efa565` reddens **three** tests:

```
pmtiles_format::header_round_trips_through_exactly_127_bytes
pmtiles_format::every_golden_header_decodes_and_re_encodes_byte_for_byte
pmtiles_reader::the_header_decodes_to_what_go_pmtiles_reports_for_every_golden
```

`the_header_decodes_to_what_go_pmtiles_reports_for_every_golden` compares all six `i32` positions
against `header.json` field for field, so the core's **decoder** is pinned properly and was all
along. The round trip catches it too, because only the read was mutated and the write still emits
little-endian, so the two stop agreeing. "A round trip cannot see endianness" is true when both ends
move together; it is not true of this mutation.

What does hold, re-measured on a clean worktree of `main` with `Compiling libviprs v0.5.0` in the
log both times:

* the leaf-entry base rebase leaves all **128** core tests green, as I reported;
* and the *writer* exchanging latitude and longitude in `set_bounds_degrees` leaves the same 128
  green, which I had not measured when I wrote the comment above. That one is the more serious of
  the pair: every archive libviprs produces would carry nonsense bounds.

So the core has two holes, just not the two I named. Filed as libviprs/libviprs#1003 with both
mutations as its acceptance criteria, and I am fixing it there rather than leaving it.

PR #215, opened 2026-09-12
Decide what the pre-push gate may skip by looking for the include_str!, not at the path

### LINKED ISSUE #214: The pre-push skip classifier calls a doc inert even when a test include_str!s it
`inert_path()` in `tools/hooks/pre-push` decides what the pre-push gate may skip by pattern alone, and one of its patterns is `docs/*`. A pattern says "this is not code". It does not say "nothing reads it", and the moment a test pulls a doc in those two stopped meaning the same thing.

libviprs PR #1010 adds `docs/pmtiles-benchmarks.md` to the core, and `tests/pmtiles_release_readiness.rs` reads it with `include_str!("../docs/pmtiles-benchmarks.md")`. Two cells assert against the contents: `the_benchmark_doc_documents_every_exported_column` and `the_benchmark_doc_names_tests_that_exist`. So a doc-only edit there genuinely breaks tests, and the gate as it stands waves that push through to fail in CI instead.

`the_skip_list_exempts_exactly_the_pinned_paths` catches it today, and its message invites me to pin the doc in `INERT_IN_LIBVIPRS` and say in the commit why nothing reads it. That route is closed here, because something does read it. Pinning it would turn a guard that is currently telling the truth into a rubber stamp.

So the classifier is what I want to fix, not the list. `inert_path()` should work the reference out for itself: scan the pushed tree's `tests/` for `include_str!` literals and treat any file one of them names as read, whichever pattern it otherwise matches. That also settles three paths the hand-written list gets wrong in the same direction: the core's own `tests/ci_feature_coverage.rs`, `tests/local_gate_is_the_job_list.rs`, `tests/doc_link_gate.rs` and `tests/miri_invocation_parity.rs` all pull in `.github/workflows/*.yml`, which the list pins as inert for a libviprs push. `docs/streaming-pdf-rotation.md`, `.gitignore` and `LICENSE` have no `include_str!` anywhere in either repo, so they stay inert on the strength of the scan rather than on the strength of a comment.

The guard has to prove the new rule rather than describe it, which means a case where a doc something `include_str!`s comes back read and a case where a doc nothing includes stays inert, both driven through the real `inert_path()` text so that deleting the scan turns them red.

spdrman:
Closes #214

`inert_path()` in `tools/hooks/pre-push` decided what the pre-push gate may skip by pattern, and `docs/*` was one of the patterns. A pattern says "this is not code". It does not say "nothing reads it", and libviprs/libviprs#1010 is where those two came apart: it adds `docs/pmtiles-benchmarks.md` and reads it from `tests/pmtiles_release_readiness.rs` with `include_str!`, with `the_benchmark_doc_documents_every_exported_column` and `the_benchmark_doc_names_tests_that_exist` asserting on the contents. A push whose only change was that doc skipped the whole suite and failed in CI instead.

`the_skip_list_exempts_exactly_the_pinned_paths` caught it, and the message it prints invited me to pin the doc in `INERT_IN_LIBVIPRS` and say in the commit why nothing reads it. That route was closed, because something does read it, and taking it would have turned a guard that was telling the truth into a rubber stamp.

So I fixed the classifier instead. For the patterns that claim nothing reads the file, `inert_path()` now scans the pushed tree's `tests/` for `include_str!` literals and treats any file one of them names as read. A literal is normalised by dropping its leading `./` and `../` steps, so it matches whether it was written from `tests/` or from a directory under it, and a tail match catches one that reached out of its own checkout. The set is built once per process, since the guard feeds every tracked path in the repo through the function.

`docs/streaming-pdf-rotation.md` stays pinned, and I checked that rather than assuming it. `git grep` across both repos finds the name in exactly one place, the pinned list in `tests/prepush_gate_cost_controls.rs` itself. Nothing includes it and nothing reads it, so it is inert on the strength of the scan now rather than on the strength of a comment.

## Proving it rather than describing it

Three cells cover the rule, and the first commit on this branch is all three going red against the classifier as it stood, naming the doc.

`a_doc_a_test_includes_is_read_and_one_nothing_includes_is_inert` builds a tree with three docs, `include_str!`s two of them (one from `tests/`, one from `tests/common/` so the answer cannot come from counting `../`) and leaves the third alone, then runs the hook's real `inert_path()` text over all three. Both halves are there because either passes on its own for the wrong reason: delete the scan and the read doc comes back inert, make the scan answer yes to everything and the unread one stops being inert. The whole-hook driver gets the same pair as two more push rows, so the end-to-end decision is pinned too.

I ran both mutations to check that:

| mutation | result |
| --- | --- |
| scan finds nothing (back to the pattern-only rule) | 3 failed, naming `docs/pmtiles-benchmarks.md` and `docs/read-by-a-test.md` |
| scan calls every candidate read | 3 failed, naming `docs/read-by-nothing.md` and the pinned `LICENSE` |

## One thing I left alone on purpose

The `.github/` and `.gitignore` entries keep their per-repo rule and never reach the scan, and that is not because the rule is right. The core's own `tests/ci_feature_coverage.rs`, `tests/doc_link_gate.rs`, `tests/local_gate_is_the_job_list.rs` and `tests/miri_invocation_parity.rs` all pull the core's workflow files in with `include_str!`, and the core's `cargo test` is half of what the pre-push suite runs, so a core workflow push does have tests that should run.

Extending the scan there cannot ride along in this PR. `.github/workflows/publish.yml` is read on the libviprs#1010 branch (`tests/pmtiles_release_readiness.rs` again) and not read at `COUNTERPART_REV`, so the pinned inert set would have to be two different sets at once: CI would go red against the counterpart, or the #1010 push would go red against the branch, which is the thing this PR exists to unblock. It needs a counterpart bump to land, and the comment in the hook and the one on `INERT_IN_LIBVIPRS` both say so.

## Evidence

Everything ran in Docker against the two checkouts laid out as siblings.

Against a detached worktree of libviprs at `origin/feat/993-pmtiles-benchmarks` (the #1010 branch), all eight cells pass:

```
running 8 tests
test run_tests_names_an_out_of_memory_kill ... ok
test a_doc_a_test_includes_is_read_and_one_nothing_includes_is_inert ... ok
test run_tests_reports_the_budget_it_will_use ... ok
test the_skip_list_never_exempts_what_the_suite_reads ... ok
test run_tests_accepts_a_fractional_memory_ceiling ... ok
test the_skip_list_exempts_exactly_the_pinned_paths ... ok
test the_skip_can_be_overridden ... ok
test the_hook_skips_and_runs_the_way_the_list_says ... ok

test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

Against the pinned counterpart (`COUNTERPART_REV` 18b7c8b4), the same eight pass and so do the six in `prepush_gate_tests_the_pushed_tree`, which shares the hook.

The real `inert_path()`, lifted out of the hook the way the guard lifts it and asked about the real #1010 core tree:

```
READ  (gate runs)   docs/pmtiles-benchmarks.md
INERT (gate skips)  docs/streaming-pdf-rotation.md
INERT (gate skips)  LICENSE
INERT (gate skips)  .gitignore
READ  (gate runs)   src/lib.rs
```

`shellcheck tools/*.sh tools/hooks/pre-push` is clean, `cargo fmt --check` passes, and `cargo clippy --all-targets -- -D warnings` is clean.


spdrman:
Merging with `Hook Mirror (every repo in the org)` red, and recording why so nobody reads that as this PR passing something it did not.

That job is already failing on `main`, and has been across the last three merges (#206, #208, #212, #213). It fails on `the_hook_mirrors_the_org_site_ci`, which complains that libviprs-org's `sync` job runs a multi-line shell block starting `rev="$(tr -d '[:space:]' < libviprs-org/cli/rust/COUNTERPART_REV)"` that the pre-commit hook does not stand in for. Nothing in this PR touches libviprs-org, that shell block, or `tools/install-hooks.sh`, so this is inherited rather than caused.

Everything this PR does touch is green: Integration Tests, Integration Tests (pdfium), Ported Cells, CLI Differential, PMTiles Interop, Lint and all four Feature Cells.

Two things worth knowing for whoever picks the Hook Mirror red up.

It is about to get harder, not easier. libviprs-org#64 adds a fifth CI job to that repo (a benchmarks drift gate), and this guard requires the hook to mirror every job in the org, so that job will need either a mirroring step or an exemption with a reason in `tools/install-hooks.sh`.

And the scan this PR adds would correctly take `.github/workflows/*.yml` off the core's inert list too, since `ci_feature_coverage.rs`, `doc_link_gate.rs`, `local_gate_is_the_job_list.rs` and `miri_invocation_parity.rs` all `include_str!` them. I deliberately did not extend it there: `publish.yml` is read on the libviprs#1010 branch and not read at the current `COUNTERPART_REV`, so the pinned set would have to be two sets at once and one side or the other would go red. That needs a counterpart bump first.

